PR #9, opened 2022-06-03
add support for uniform structured grids

### LINKED ISSUE #8: adding support for Structured grid format?
Hi, thanks for your work on ReadVTK.jl, it seems a very useful tool with high potential. I am using WriteVTK.jl in some projects and at some near point I will need to read vtk files as well.

I see that your package does not support the structured grid format. Have you considered adding that? I have some matlab vtk-filereading functions and before migrating them to a simple julia version I would like to consider contributing to this...

Thanks, Jorge

jorgepz:
closes #8 

for debugging I am using this MWE

```julia
# ReadVTK.jl gridExample
using WriteVTK, ReadVTK


x, y, z = 0:10, 1:6, 2:0.1:3
Nx, Ny, Nz = length(x), length(y), length(z)

scalarField = rand(Nx, Ny, Nz)

vtk_grid("grid", x, y, z) do vtk
    vtk["Temperature", VTKPointData()] = scalarField              # scalar field attached to points
end

filepath = "grid.vti"
vtk = VTKFile( filepath )
```


jorgepz:
I have a first MWE where cell data is well read and a validation is done. the changes in readvtk are intentionally minimal in order to avoid conflicts. the test example i am running (and by the moment I did not include in the repo) is the following 

```julia

# ReadVTK.jl gridExample
using WriteVTK, ReadVTK

## Generate grid file and write vti
x, y, z = 1:3, 1:2, 2:0.1:2.2
Nx, Ny, Nz = length(x), length(y), length(z)

pointScalarField = rand(Nx, Ny, Nz)
cellScalarField  = rand(Nx-1, Ny-1, Nz-1)

print( "cell scalar field", cellScalarField,"\n")

cellDataName  = "Name of Test Cell scalar data"
pointDataName = "Point scalar data"

print("  writing vtk file...")
vtk_grid("grid", x, y, z) do vtk
    vtk[ pointDataName, VTKPointData()] = pointScalarField  # scalar field attached to points
    vtk[ cellDataName, VTKCellData()] = cellScalarField     # scalar field attached to cells
end
print("done.\n")

## Read vti file
print("  reading vtk file...")
filepath = "grid.vti"
vtk = VTKFile( filepath )
data = get_data( get_cell_data(vtk)[cellDataName] )
print("done.\n")

reshapedData = reshape( cellScalarField, ( (Nx-1), (Ny-1), (Nz-1) ) )

difference = reshapedData .- cellScalarField
```

ranocha:
Thanks a lot! I didn't dive through the code, but I would be fine with adding WriteVTK.jl as test dependency. This would make it much simpler to extend things in the future and make sure that the Julia VTK packages are compatible with each other.

sloede:
> but I would be fine with adding WriteVTK.jl as test dependenc

Absolutely! It would be great if you can add some tests, maybe along the lines of what you gave as an MWE.

The implementation looks good so far as well. Some things to consider:
* Should we maybe add functions like `get_extent` and/or `get_origin`?
* Should we return the data as plain vectors (as we are now), or should we return them as multi-dimensional arrays?
* If returning as vector, maybe it would be nice to have a function like `size` that accepts a VTKDataArray (or VTKData, haven't really thought about it yet), that can be used to reshape the returned vector in its original size?

jorgepz:
Hi, thanks both for your comments!

In [c75860d](https://github.com/trixi-framework/ReadVTK.jl/pull/9/commits/c75860ddacb4867aafdf1132054685bd979dacf3) I included my small test script in the runtest.jl script and it seems to be working. I also added the dependency to the tests Toml as I @ranocha suggested, however I am not experienced (I am a Julia newbie) in using diifferent toml's, so in order to avoid an error I am activating from the root project and including the runtest.jl from there (with WriteVTK) installed. A documentation entry for the right approach for the testing would be nice too I think, but I would not do that in this PR to focus. By the moment I am in my root folder opening julia with `julia --project=.` and from there `using Revise` and `include("test/runtests.jl")`. That works well (if I have installed writevtk).

Regarding the questions from @sloede 
 - I think that this depends on how compatible ReadVTK is supposed to be with WriteVTK... if the answer is "a lot" then I think that the most natural would be to return the data in the same format that WriteVTK introduces it to the vtk... thus a multidimensional array. 
 - Regarding the potential `get_origin` function, it is indeed needed in order to fully read the grid information. I could make a first try, probably in another .jl file...
 - the same happens with the sizes, I think that this should also be read from the vti file, or if the data array is multidimensional, then you can already get that info as well...

jorgepz:
In [3e7683a](https://github.com/trixi-framework/ReadVTK.jl/pull/9/commits/3e7683af1707e4d68f937ce7218236b52bf9febb) I added a small get_origin function and also a test for that. I also separated that in another file, as well as moved the exports to another file.. please let me know if you prefer to keep everything in a single file... if you agree on this approach I could continue with the other fields like: extent, spacing and point data.

ranocha:
> in order to avoid an error I am activating from the root project and including the runtest.jl from there (with WriteVTK) installed. A documentation entry for the right approach for the testing would be nice too I think, but I would not do that in this PR to focus. By the moment I am in my root folder opening julia with `julia --project=.` and from there `using Revise` and `include("test/runtests.jl")`. That works well (if I have installed writevtk).

That sounds reasonable. Alternatively, you could also activate the main environment of ReadVTK.jl and run
```julia
using Pkg; Pkg.test("ReadVTK")
```
to run all tests of ReadVTK.jl. Then, the Julia package manager will take care of installing WriteVTK (in a temporary environment for running the tests).

src/ReadVTK.jl
```diff
@@ -8,10 +8,6 @@ using CodecZlib: ZlibDecompressor
 using LightXML: LightXML, XMLDocument, XMLElement, parse_string, attribute, has_attribute,
                 child_elements, free, content, find_element
 
-export VTKFile, get_points, get_cells, VTKData, get_point_data, get_cell_data, VTKDataArray,
-       get_data, VTKCells, get_example_file
```
ranocha, line 12: I would prefer to keep the `export`s in our main file. This is the canonical approach used by most Julia packages I know and allows people to easily find all `export`ed symbols (which often form the public API of a Julia package).

src/get_functions.jl
```diff
@@ -0,0 +1,19 @@
+
+"""
+get_origin(vtk_file)
+
+Retrieve the origin of a VTK regular structured grid file.
+
+"""
+
+function get_origin( vtk_file )
```
ranocha, line 9:
```suggestion
"""
function get_origin(vtk_file)
```
Otherwise, the docstring is not attached to the function.

src/get_functions.jl
```diff
@@ -0,0 +1,19 @@
+
+"""
+get_origin(vtk_file)
```
ranocha, line 3:
```suggestion
    get_origin(vtk_file)
```
To get correct rendering as code

test/Project.toml
```diff
@@ -1,2 +1,3 @@
 [deps]
 Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
+WriteVTK = "64499a7a-5c06-52f2-abe2-ccb03c286192"
```
ranocha, line 3: Could you please add a `[compat]` section with a (lower) bound on WriteVTK.jl? See https://pkgdocs.julialang.org/v1/compatibility

test/runtests.jl
```diff
@@ -163,6 +163,44 @@ mkpath(TEST_EXAMPLES_DIR)
...
+    ## Generate grid file and write vti
+    inputOrigin = [1.0, 1.0, 2.0]
+    x, y, z = (inputOrigin[1]):3, (inputOrigin[2]):2, (inputOrigin[3]):0.1:2.2
+    Nx, Ny, Nz = length(x), length(y), length(z)
+
+    pointScalarField = rand(Nx, Ny, Nz)
+    cellScalarField  = rand(Nx-1, Ny-1, Nz-1)
+
+    cellDataName  = "Name of Test Cell scalar data"
+    pointDataName = "Point scalar data"
```
ranocha, line 181: Would you mind following the common Julia style to use `snake_case` for functions and variables? See https://docs.julialang.org/en/v1/manual/style-guide/#Use-naming-conventions-consistent-with-Julia-base/

ranocha (changes requested):
Thanks a lot! Please find below some comments. I will trigger a first CI run.

src/ReadVTK.jl
jorgepz: ok, thanks. restored in [198cd6e](https://github.com/trixi-framework/ReadVTK.jl/pull/9/commits/198cd6e68edcc3562bf15760c5055f8214495f87)

src/ReadVTK.jl
```diff
@@ -126,16 +123,34 @@ function VTKFile(filename)
...
+    n_cells = prod( n_points_per_grid_dir .- 1 )
+
+    # Create and return VTKFile
+    VTKFile(filename, xml_file, file_type, version, byte_order, compressor, appended_data, n_points, n_cells)
+
+  end
 
-  # Create and return VTKFile
-  VTKFile(filename, xml_file, file_type, version, byte_order, compressor, appended_data, n_points,
-          n_cells)
```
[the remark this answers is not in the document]
jorgepz: Indeed. Thanks for that, I was fixed in [bb8b250](https://github.com/trixi-framework/ReadVTK.jl/pull/9/commits/bb8b2502f0555072b95ea24210f1b30d94e02217) . the first version was considering a different call to VTKfile but in the end I was able (and preferred) to keep it the same

src/get_functions.jl
jorgepz: I was not writing yet the docs but is good to know that, thanks!

test/Project.toml
jorgepz: added in [c4a1ea7](https://github.com/trixi-framework/ReadVTK.jl/pull/9/commits/c4a1ea7cdcf899dfe6ee4e6c452f4471fb8486c0) thanks for the reference to the docs (I never included compats). I used version 1.13 as a suggestion, and the tests work fine, but I guess there might be a lower bound that works.

test/runtests.jl
jorgepz: Thanks. I think this is now solved in [b11af19](https://github.com/trixi-framework/ReadVTK.jl/pull/9/commits/b11af19d2980230e6b52b88c749ccc575b2779a7)

src/ReadVTK.jl
```diff
@@ -109,8 +106,8 @@ function VTKFile(filename)
     compressor = ""
   end
 
-  # Ensure matching file type
-  @assert file_type == "UnstructuredGrid"
+  # Ensure matching file types
+  @assert ( file_type == "UnstructuredGrid" || file_type == "ImageData" )
```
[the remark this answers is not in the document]
sloede: Yes, I agree. It was my fault for doing it this way, but it would be great if you could change it @jorgepz.

If possible, I'd prefer to change this to something like
```julia
if !(file_type in ("UnstructuredGrid", "ImageData"))
  error("Unsupported file type: ", file_type)
end
```

test/Project.toml
sloede: That's fine, thanks!

test/runtests.jl
```diff
@@ -163,6 +163,44 @@ mkpath(TEST_EXAMPLES_DIR)
...
+    origin = get_origin( vtk )
+    data   = get_data( get_cell_data(vtk)[cellDataName] )
+    
+    reshapedData = reshape( cellScalarField, ( (Nx-1), (Ny-1), (Nz-1) ) )
+
+    # test if cell data is well read
+    @test iszero( reshapedData .- cellScalarField )
+
+    # test if the origin is well read
+    @test iszero( origin .- inputOrigin )
```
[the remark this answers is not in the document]
sloede: I agree that
```julia
@test a_written == a_read
```
is much clearer and that's what I would prefer as well.

Re accuracy: Since we are reading and writing *unmodified* data, we expect binary identity as otherwise we did something wrong. Thus I'd prefer to use exact equality in this test as suggested by @jorgepz and only consider FP accuracy if we have unforeseen (and unfixable) issues with that.

src/ReadVTK.jl
```diff
@@ -126,16 +128,27 @@ function VTKFile(filename)
...
-  n_cells = parse(Int, attribute(piece, "NumberOfCells", required=true))
+  if file_type == "UnstructuredGrid"
+    # Extract piece
+    piece = root[file_type][1]["Piece"][1]
+    n_points = parse(Int, attribute(piece, "NumberOfPoints", required=true))
+    n_cells = parse(Int, attribute(piece, "NumberOfCells", required=true))
+
+  elseif file_type == "ImageData"
+    piece = root[file_type][1]
+    wholeExtent = parse.(Int,     split( attribute(piece, "WholeExtent", required=true) , ' ' ) )
```
sloede, line 141: What you are reading in here is not a "piece" in VTK parlance, but a "dataset element", thus I suggest to change it to
```suggestion
    dataset_element = root[file_type][1]
    wholeExtent = parse.(Int, split(attribute(dataset_element, "WholeExtent", required=true), ' '))
```
I also removed unnecessary whitespace

src/ReadVTK.jl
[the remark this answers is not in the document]
jorgepz: No problem. Done in [512abf8](https://github.com/trixi-framework/ReadVTK.jl/pull/9/commits/512abf8bbe49c3af6f28ff66bda5b9b88dc57da2)

src/ReadVTK.jl
```diff
@@ -126,16 +128,27 @@ function VTKFile(filename)
...
+    n_points_per_grid_dir = [ wholeExtent[2*i]+1 for i in (1:3) ]
+    n_points = prod( n_points_per_grid_dir )
+    n_cells = prod( n_points_per_grid_dir .- 1 )
```
sloede, line 144:
```suggestion
    n_points_per_grid_dir = [wholeExtent[2*i]+1 for i in (1:3)]
    n_points = prod(n_points_per_grid_dir)
    n_cells = prod(n_points_per_grid_dir .- 1)
```
We generally do not use a space inside parens/brackets. Would you mind changing this also elsewhere to keep the formatting style consistent with the existing code? It seems trivial, but in our experience it greatly helps in reading code written by someone else if everything is formatted consistently.

test/runtests.jl
```diff
@@ -163,6 +163,54 @@ mkpath(TEST_EXAMPLES_DIR)
       @test sum(data) ≈ 192.1204941112099
     end
   end
+
+  # ----------------------------
+  # Test for validation of structured grid read feature
+  # ----------------------------
```
sloede, line 169:
```suggestion
  # Test for validation of uniform grid ("image data") read feature
```
Just to be clear that we do not have support for general structured meshes yet

src/get_functions.jl
```diff
@@ -0,0 +1,40 @@
+"""
+    get_origin( vtk_file )
+
+Retrieve the vector of coordinates of the origin of a regular structured grid from a given VTK XML file.
+
+See also: [`VTKFile`](@ref)
+"""
+function get_origin( vtk_file )
+
```
sloede, line 9:
```suggestion
```
No leading empty lines (please also change in other parts of this PR)

src/get_functions.jl
```diff
@@ -0,0 +1,40 @@
...
+  return origin_point
+end
+
+
+
+"""
+    get_spacing( vtk_file )
+
+Retrieve a vector with the regular increments in each coordinate of the regular structured grid from a given VTK XML file.
```
sloede, line 25:
```suggestion
Retrieve a vector with the regular increments in each coordinate direction of the uniform grid from the given [`VTKFile`](@ref) file.
```

src/get_functions.jl
```diff
@@ -0,0 +1,40 @@
...
+
+Retrieve the vector of coordinates of the origin of a regular structured grid from a given VTK XML file.
+
+See also: [`VTKFile`](@ref)
+"""
+function get_origin( vtk_file )
+
+  # open the file and locate the ImageData section
+  root = LightXML.root( vtk_file.xml_file )
+  piece = root["ImageData"][1]
```
sloede, line 12:
```suggestion
  dataset_element = root["ImageData"][1]
```
As above. Also applies to `get_spacing` below.

src/get_functions.jl
```diff
@@ -0,0 +1,40 @@
+"""
+    get_origin( vtk_file )
+
+Retrieve the vector of coordinates of the origin of a regular structured grid from a given VTK XML file.
```
sloede, line 4:
```suggestion
Retrieve the vector of coordinates of the origin of a uniform grid from the given [`VTKFile`](@ref) file.
```

src/get_functions.jl
```diff
@@ -0,0 +1,40 @@
...
+
+See also: [`VTKFile`](@ref)
+"""
+function get_origin( vtk_file )
+
+  # open the file and locate the ImageData section
```
sloede, line 10: As a suggestion/question: Should we check if `vtk_file.filetype == "ImageData"` and throw an error otherwise to inform users if they misuse this function on an unstructured mesh? Same applies to `get_spacing`.

src/get_functions.jl
```diff
@@ -0,0 +1,40 @@
...
+  root = LightXML.root( vtk_file.xml_file )
+  piece = root["ImageData"][1]
+
+  # obtain the origin
+  origin_point = parse.(Float64, split( attribute(piece, "Origin", required=true) , ' ' ) )
+
+  return origin_point
```
sloede, line 17: General question: Should we always return a 3-dimensional vector even if it is a 2D file?

It's what VTK does internally, but as a user I'd find it highly unusual. I'd also be interested in your take on this @ranocha.

sloede (commented):
I did an initial review of the PR, and overall it is shaping up nicely! Thanks a lot for the contribution @jorgepz!

I left a few comments/suggestions. Regarding capabilities, I was wondering:
* Have you tried using it for 2D files as well? Should we maybe add a test for a 2D file?
* Have you tried using it for non-scalar datasets, e.g., vectorial or tensor fields?

src/ReadVTK.jl
jorgepz: excellent. thanks.

test/runtests.jl
jorgepz: good clarification

src/get_functions.jl
jorgepz: yes. Thought about that for a future PR, but I already added it now.

src/get_functions.jl
jorgepz: did not see the commit suggestion and already did it in [e3fe8e1](https://github.com/trixi-framework/ReadVTK.jl/pull/9/commits/e3fe8e1728a33377d6e8f2a471c5b214f24005c6)

src/get_functions.jl
jorgepz: Thanks for your comments @sloede!

>     * Have you tried using it for 2D files as well? Should we maybe add a test for a 2D file?

I did not try to use 2D files, since in the applications I am planning to use ReadVTK and WriteVTK everything is computed in a general 3D manner (even for 2D or 1D grids). I do not know the interval vtk structure to know if it stores data in a different way. I would guess that it does not.

>     * Have you tried using it for non-scalar datasets, e.g., vectorial or tensor fields?

This is a useful feature that I will be needing in the future, and I would definitely add a test for this in the following PR. By the moment I thought about starting with scalar and see how I went...

To have in mind the applications I am planning to solve: I'll be using WriteVTK and ReadVTK to generate images or 3D solids during deformation to solve these kind of problems https://www.sciencedirect.com/science/article/abs/pii/S0045782519302099

jorgepz:
Thanks for your comments @sloede!

>   * Have you tried using it for 2D files as well? Should we maybe add a test for a 2D file?

I did not try to use 2D files, since in the applications I am planning to use ReadVTK and WriteVTK everything is computed in a general 3D manner (even for 2D or 1D grids). I do not know the interval vtk structure to know if it stores data in a different way. I would guess that it does not.

>    * Have you tried using it for non-scalar datasets, e.g., vectorial or tensor fields?

This is a useful feature that I will be needing in the future, and I would definitely add a test for this in the following PR. By the moment I thought about starting with scalar and see how I went...

To have in mind the applications I am planning to solve: I'll be using WriteVTK and ReadVTK to generate images or 3D solids during deformation to solve these kind of problems https://www.sciencedirect.com/science/article/abs/pii/S0045782519302099

src/ReadVTK.jl
```diff
@@ -126,10 +130,19 @@ function VTKFile(filename)
...
-  # Extract piece
-  piece = root[file_type][1]["Piece"][1]
-  n_points = parse(Int, attribute(piece, "NumberOfPoints", required=true))
-  n_cells = parse(Int, attribute(piece, "NumberOfCells", required=true))
+  if file_type == "UnstructuredGrid"
+    # Extract piece
+    piece = root[file_type][1]["Piece"][1]
+    n_points = parse(Int, attribute(piece, "NumberOfPoints", required=true))
+    n_cells = parse(Int, attribute(piece, "NumberOfCells", required=true))
+
```
sloede, line 138:
```suggestion
```
unnecessary empty line

src/get_functions.jl
```diff
@@ -0,0 +1,48 @@
+"""
+    get_origin( vtk_file )
```
sloede, line 2:
```suggestion
    get_origin(vtk_file)
```
unnecessary whitespace

src/get_functions.jl
```diff
@@ -0,0 +1,48 @@
...
+  # obtain the origin
+  origin_point = parse.(Float64, split(attribute(dataset_element, "Origin", required=true), ' '))
+
+  return origin_point
+end
+
+
+
+"""
+    get_spacing( vtk_file )
```
sloede, line 27:
```suggestion
    get_spacing(vtk_file)
```
unnecessary whitespace

src/get_functions.jl
```diff
@@ -0,0 +1,48 @@
+"""
+    get_origin( vtk_file )
+
+Retrieve the vector of coordinates of the origin of a uniform grid from the given [`VTKFile`](@ref) file.
+
+See also: [`VTKFile`](@ref)
+"""
+function get_origin(vtk_file)
+  # check imagedata
+  if !(vtk_file.file_type == "ImageData")
```
sloede, line 10:
```suggestion
  if vtk_file.file_type != "ImageData"
```
easier to comprehend

src/get_functions.jl
```diff
@@ -0,0 +1,48 @@
...
+  end
+
+  # open the file and locate the ImageData section
+  root = LightXML.root(vtk_file.xml_file)
+  dataset_element = root["ImageData"][1]
+
+  # obtain the origin
+  origin_point = parse.(Float64, split(attribute(dataset_element, "Origin", required=true), ' '))
+
+  return origin_point
```
sloede, line 21: Minor detail: I'd just call the return value `origin`, given that the function is named `get_origin`

test/runtests.jl
```diff
@@ -163,6 +163,55 @@ mkpath(TEST_EXAMPLES_DIR)
       @test sum(data) ≈ 192.1204941112099
     end
   end
+
+  # Test for validation of uniform grid ("image data") read feature
+  @testset "ImageData" begin
+
```
sloede, line 169:
```suggestion
```
unnecessary empty line

test/runtests.jl
```diff
@@ -163,6 +163,55 @@ mkpath(TEST_EXAMPLES_DIR)
...
+    point_data_read = reshape( point_data_read,  (     Nx,     Ny,     Nz ) )
+
+    # test if cell and point data are well read
+    @test cell_data_read == cell_scalar_field
+    @test point_data_read == point_scalar_field
+
+    # test if the origin and spacing are well read
+    @test origin  == input_origin
+    @test spacing == input_spacing
+
```
sloede, line 213:
```suggestion
```

test/runtests.jl
```diff
@@ -163,6 +163,55 @@ mkpath(TEST_EXAMPLES_DIR)
       @test sum(data) ≈ 192.1204941112099
     end
   end
+
+  # Test for validation of uniform grid ("image data") read feature
+  @testset "ImageData" begin
```
sloede, line 168: Would you mind also removing the whitespace inside parens/brackets in this file?

test/runtests.jl
```diff
@@ -163,6 +163,55 @@ mkpath(TEST_EXAMPLES_DIR)
...
+    cell_data_read = reshape( cell_data_read,    ( (Nx-1), (Ny-1), (Nz-1) ) )
+    point_data_read = reshape( point_data_read,  (     Nx,     Ny,     Nz ) )
+
+    # test if cell and point data are well read
+    @test cell_data_read == cell_scalar_field
+    @test point_data_read == point_scalar_field
+
+    # test if the origin and spacing are well read
+    @test origin  == input_origin
+    @test spacing == input_spacing
```
sloede, line 212: I would slightly re-arrange the tests to make it easier to spot where the problem is in case of an error:
```suggestion
    # Read vti file using ReadVTK
    filepath = "grid.vti"
    @testset "VTKFile" begin
      @test VTKFile(filepath) isa VTKFile
    end
    vtk = VTKFile(filepath)

    # check getter functions
    @testset "get_origin" begin
      @test get_origin(vtk) == input_origin
    end

    @testset "get_spacing" begin
      @test get_spacing(vtk) == input_spacing
    end

    @testset "get cell data" begin
      cell_data_raw = get_data(get_cell_data(vtk)[cell_data_name])
      cell_data_reshaped = reshape(cell_data_raw, ((Nx-1), (Ny-1), (Nz-1)))
      @test cell_data_reshaped == cell_scalar_field
    end

    @testset "get point data" begin
      point_data_raw = get_data(get_point_data(vtk)[point_data_name])
      point_data_reshaped = reshape(point_data_raw, (Nx, Ny, Nz))
      @test point_data_reshaped == point_scalar_field
    end
```

test/runtests.jl
jorgepz: done in [69ac557](https://github.com/trixi-framework/ReadVTK.jl/pull/9/commits/69ac5572b9d8d2ca5299a3d21c36da0c54c0a86c)

src/ReadVTK.jl
```diff
@@ -8,9 +8,10 @@ using CodecZlib: ZlibDecompressor
...
-export VTKFile, get_points, get_cells, VTKData, get_point_data, get_cell_data, VTKDataArray,
-       get_data, VTKCells, get_example_file
-
+export VTKFile, VTKData, VTKDataArray, VTKCells,       # structs
+       get_point_data, get_cell_data, get_data,        # get data functions
+       get_points, get_cells, get_origin, get_spacing, # get geometry functions
+       get_example_file                                # other functions
 
 """
     VTKFile
```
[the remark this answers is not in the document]
jorgepz: done in [b6f834c](https://github.com/trixi-framework/ReadVTK.jl/pull/9/commits/b6f834c1aa08e3ae7043961a598b85da6b0936f3)

src/get_functions.jl
jorgepz: done in [2d1f3dc](https://github.com/trixi-framework/ReadVTK.jl/pull/9/commits/2d1f3dcdf69439071136b6f53f34fa803ffc2579)

src/get_functions.jl
jorgepz: I did a small test with
```julia

## Generate grid file and write vti
x, y = 1:3, 1:2
Nx, Ny = length(x), length(y)

pointScalarField = rand(Nx, Ny)
cellScalarField  = rand(Nx-1, Ny-1)

print( "cell scalar field", cellScalarField,"\n")

cellDataName  = "Name of Test Cell scalar data"
pointDataName = "Point scalar data"

vtk_grid("grid", x, y) do vtk
    vtk[ pointDataName, VTKPointData()] = pointScalarField  # scalar field attached to points
    vtk[ cellDataName, VTKCellData()] = cellScalarField     # scalar field attached to cells
end
print("done.\n")

## Read vti file
print("  reading vtk file...")
filepath = "grid.vti"
vtk = VTKFile( filepath )
data = get_data( get_cell_data(vtk)[cellDataName] )
print("done.\n")
```
it works and generates a 2D vti that is opened by paraview as plane but... the data in the header is saved with 3 components with extent "0 0" in the z direction... thus it seems that detecting if the problem is 2D or not requires only to check the extent of z... if it is useful in the entire trixi framework then it could be added...

sloede: Thanks for checking this out!

I am trying to think about this question from a user perspective: What would a user expect? If we look at it from the WriteVTK point of view, I think as a user I'd want to get back the same information that I put in. Since when creating a 2D file I only provide 2D information, I think we should return 2D information in ReadVTK as well.

What's your take on this?

jorgepz: I agree. I would also suggest going with an approach coherent with what writevtk reads and allows the user to do... so I would "read" 2D origins and as well read 3D point data arrays...

src/ReadVTK.jl
```diff
@@ -126,10 +130,18 @@ function VTKFile(filename)
...
-  n_cells = parse(Int, attribute(piece, "NumberOfCells", required=true))
+  if file_type == "UnstructuredGrid"
+    # Extract piece
+    piece = root[file_type][1]["Piece"][1]
+    n_points = parse(Int, attribute(piece, "NumberOfPoints", required=true))
+    n_cells = parse(Int, attribute(piece, "NumberOfCells", required=true))
+  elseif file_type == "ImageData"
+    dataset_element = root[file_type][1]
+    wholeExtent = parse.(Int, split(attribute(dataset_element, "WholeExtent", required=true), ' '))
+    n_points_per_grid_dir = [wholeExtent[2*i]+1 for i in (1:3)]
```
sloede, line 141:
```suggestion
    whole_extent = parse.(Int, split(attribute(dataset_element, "WholeExtent", required=true), ' '))
    n_points_per_grid_dir = [whole_extent[2*i]+1 for i in (1:3)]
```
Variables/functions should be snake case

src/get_functions.jl
```diff
@@ -0,0 +1,48 @@
...
+  # check imagedata
+  if vtk_file.file_type != "ImageData"
+    error("spacing can be read only for ImageData file_type.")
+  end
+  
+  # open the file and locate the ImageData section
+  root = LightXML.root(vtk_file.xml_file)
+  dataset_element = root["ImageData"][1]
+
+  # obtain the origin
```
sloede, line 43:
```suggestion
  # obtain the spacing
```

src/get_functions.jl
sloede: Would you mind modifying the `get_origin` and `get_spacing` functions accordingly, i.e., such that for 2D files it only returns a two-component vector?

sloede (changes requested):
Overall, this looks good to me! Once `get_spacing`/`get_origin` return two-component vectors for 2D files, I think this can be merged. I also left two minor suggestions.

Thanks for all the good work!

src/get_functions.jl
ranocha: > I am trying to think about this question from a user perspective: What would a user expect? If we look at it from the WriteVTK point of view, I think as a user I'd want to get back the same information that I put in. Since when creating a 2D file I only provide 2D information, I think we should return 2D information in ReadVTK as well.

Yes, that sounds reasonable to me :+1:

ranocha (commented):
Thanks a lot! Everything is fine from my point of view when the remaining comments of @sloede are resolved.

sloede:
One further note @jorgepz: I think you should add your name to the authors' list. For example, you could add
```markdown
Further contributions to ReadVTK have been made by the following people:
* Jorge Pérez Zerpa
```
after https://github.com/jorgepz/ReadVTK.jl/blob/fae422c9d4f126835a8ca9cf4f77903b6b35ca4e/README.md?plain=1#L103

and
https://github.com/jorgepz/ReadVTK.jl/blob/fae422c9d4f126835a8ca9cf4f77903b6b35ca4e/docs/src/index.md?plain=1#L97

src/get_functions.jl
jorgepz: Thanks both for your comments. I think there is something interesting here. I am trying to understand vtks differences in both cases... I did this test
```julia

## Generate grid file and write vti
x, y, z = 1:3, 1:2, 0:0
Nx, Ny, Nz = length(x), length(y), length(z)

pointScalarField2D = rand(Nx, Ny )
cellScalarField2D  = rand(Nx-1, Ny-1 )

print( "cell scalar field 2D", cellScalarField2D,"\n")

pointScalarField3D = rand(Nx, Ny, Nz)
cellScalarField3D  = rand(Nx-1, Ny-1, Nz-1)

print( "cell scalar field 3D", cellScalarField3D,"\n")

cellDataName  = "Cell scalar data label"
pointDataName = "Point scalar data"

vtk_grid("grid_2D", x, y) do vtk
    vtk[ pointDataName, VTKPointData()] = pointScalarField2D  # scalar field attached to points
    vtk[ cellDataName, VTKCellData()] = cellScalarField2D     # scalar field attached to cells
end

vtk_grid("grid_3D", x, y,z) do vtk
    vtk[ pointDataName, VTKPointData()] = pointScalarField3D  # scalar field attached to points
    vtk[ cellDataName, VTKCellData()] = cellScalarField3D     # scalar field attached to cells
end

## Read vti file
filepath = "grid_2D.vti"
vtk = VTKFile( filepath )
data = get_data( get_point_data(vtk)[pointDataName] )
data = get_data( get_cell_data(vtk)[cellDataName] )
print( get_origin( vtk ) )

filepath = "grid_3D.vti"
vtk = VTKFile( filepath )
data = get_data( get_point_data(vtk)[pointDataName] )
data = get_data( get_cell_data(vtk)[cellDataName] )
print( get_origin( vtk ) )
```

and, the question is how do we detect if a files is 2D or 3D?  The 2D vti file can be well opened in paraview and it seems ok and the 3D file as expected has an error. apart from that, the number of components in the 3D is zero... that should be because we wrote a scalar magnitude in a zero z extent mesh...

the content of the grid2D is here
```
<?xml version="1.0" encoding="utf-8"?>
<VTKFile type="ImageData" version="1.0" byte_order="LittleEndian" header_type="UInt64" compressor="vtkZLibDataCompressor">
  <ImageData WholeExtent="0 2 0 1 0 0" Origin="1 1 0" Spacing="1 1 1">
    <Piece Extent="0 2 0 1 0 0">
      <PointData>
        <DataArray type="Float64" Name="Point scalar data" NumberOfComponents="1" format="appended" offset="0"/>
      </PointData>
      <CellData>
        <DataArray type="Float64" Name="Cell scalar data label" NumberOfComponents="1" format="appended" offset="91"/>
      </CellData>
    </Piece>
  </ImageData>
  <AppendedData encoding="raw">
_
```

the (copy-paste) content of the grid3D is here
```
<?xml version="1.0" encoding="utf-8"?>
<VTKFile type="ImageData" version="1.0" byte_order="LittleEndian" header_type="UInt64" compressor="vtkZLibDataCompressor">
  <ImageData WholeExtent="0 2 0 1 0 0" Origin="1 1 0" Spacing="1 1 1">
    <Piece Extent="0 2 0 1 0 0">
      <PointData>
        <DataArray type="Float64" Name="Point scalar data" NumberOfComponents="1" format="appended" offset="0"/>
      </PointData>
      <CellData>
        <DataArray type="Float64" Name="Cell scalar data label" NumberOfComponents="0" format="appended" offset="91"/>
      </CellData>
    </Piece>
  </ImageData>
  <AppendedData encoding="raw">
_
```

**bottom line**: 2D files (x-y) can be identified by the number of extent in z. I will try to add that.

jorgepz: I think this is solved in [105b5b0](https://github.com/trixi-framework/ReadVTK.jl/pull/9/commits/105b5b09fa997008145f175d6a3cf0e941c4dc48)

jorgepz:
> One further note @jorgepz: I think you should add your name to the authors' list. For example, you could add

Thanks @sloede . I did a proposal in [d95f37b](https://github.com/trixi-framework/ReadVTK.jl/pull/9/commits/d95f37b512ea2176acd6f2c4b9098c077492351e)

src/get_functions.jl
```diff
@@ -0,0 +1,88 @@
...
+Retrieve a vector with the extent in each direction of the uniform grid from the given [`VTKFile`](@ref) file.
+
+See also: [`VTKFile`](@ref)
+"""
+function get_extent(vtk_file)
+
+  # obtain dataset
+  dataset_element = get_imagedata_dataset(vtk_file)
+
+  # obtain the origin
```
sloede, line 62:
```suggestion
  # obtain extent
```

sloede (changes requested):

src/get_functions.jl
```diff
@@ -0,0 +1,88 @@
...
+  extent = parse.(Float64, split(attribute(dataset_element, "WholeExtent", required=true), ' '))
+
+  return extent
```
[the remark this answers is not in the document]
jorgepz: I think that guessing if a vti file is 2D or 3D will not be as straightforward as one might think. By what I understand WholeExtent is always formed by 6 integers. In the case of 3D the last are 0 and something higher than zero, and in the 2D case they should be both be zero. I renamed the extent to whole_extent, so that we read exactly what vtk writes and also I added a zero z component verification to provide a 2D origin.

src/get_functions.jl
jorgepz: yes. I had to reclone the repo and remake some changes. now it's ok. thanks.

sloede: For the future: I recommend to *not* work in `main` of a fork, but to create a separate branch in your fork on which you do your changes. This will make it much easier to synchronize your fork with the upstream repo.

jorgepz: thanks!

src/get_functions.jl
```diff
@@ -0,0 +1,83 @@
...
+function get_origin(vtk_file)
+  # obtain dataset
+  dataset_element = get_imagedata_dataset(vtk_file)
+
+  whole_extent = get_whole_extent(vtk_file)
+  
+  # obtain the origin
+  origin = parse.(Float64, split(attribute(dataset_element, "Origin", required=true), ' '))
+
+  if (whole_extent[5:6] == [0,0]) && (origin[3] == 0)
```
sloede, line 17:
```suggestion
  if (whole_extent[5:6] == [0,0]) && (origin[3] == 0.0)
```
`origin` is a `Vector{Float64}`

src/get_functions.jl
```diff
@@ -0,0 +1,83 @@
...
+Retrieve a vector with the `WholeExtent` 6-entry vector from the uniform grid [`VTKFile`](@ref) file.
+
+See also: [`VTKFile`](@ref)
+"""
+function get_whole_extent(vtk_file)
+  # obtain dataset
+  dataset_element = get_imagedata_dataset(vtk_file)
+
+  # obtain extent
+  whole_extent = parse.(Float64, split(attribute(dataset_element, "WholeExtent", required=true), ' '))
```
sloede, line 59:
```suggestion
  whole_extent = parse.(Int, split(attribute(dataset_element, "WholeExtent", required=true), ' '))
```
The extent is always an integer I believe.

sloede (approved):
LGTM from my side! If all tests pass, I think this can be merged. I did a few manual changes @jorgepz in order not to bug you with more details, I hope that's OK. Thanks a lot for your contribution!

sloede:
@jorgepz Unfortunately, there is an error in the Windows tests: https://github.com/trixi-framework/ReadVTK.jl/runs/6833929733?check_suite_focus=true

jorgepz:
Yes thanks @sloede  it seems it was working here https://github.com/trixi-framework/ReadVTK.jl/runs/6833787929?check_suite_focus=true ... will check later. thanks.

sloede:
The Windows error disappeared after retriggering the tests. Let's hope that they stay away.

PR #11, opened 2022-06-24
PolyData

### LINKED ISSUE #10: Support for PolyData?
Would you please consider adding PolyData support? It would be greatly appreciated. I use WriteVTK to save results from particle simulations and then view in paraview. Sometimes I would like to load a simulation state or do some complicated post-processing.  
I have only bunch of points with point data. Actually, your package works when I add
```
  elseif file_type == "PolyData"
    piece = root[file_type][1]["Piece"][1]
    n_points = parse(Int, attribute(piece, "NumberOfPoints", required=true))
    n_cells = 0
  end 
```
after line 140. But by default it throws assertion error.

OndrejKincl:
As requested, I send a very simple modification which allows to handle polydata and a test which shows that it works.

Disclaimers:
- I did not know what to set as n_cells. Polydata does not really have cells. I set it to zero.
- Be warned that I am a mathematician. Not a programmer.
- Polydata also allow lines and triangle patches which I did not test.

Closes #10

### DIFF
--- a/README.md
+++ b/README.md
@@ -74,16 +74,18 @@ Further example VTK files can be found in the
 * Extracting point coordinates
 * Extracting information about cell types
 * Only for `ImageData` files: get origin, spacing, and extent information
+* Reading `PolyData` files containing vortices, lines and/or polygons
 
 ### What does not work
 * Reading VTK files not stored in the VTK XML format
-* Reading VTK files of other type than `UnstructuredGrid`/`ImageData`
+* Reading VTK files of other type than `UnstructuredGrid`, `ImageData` or `PolyData`
 * Multiblock files, PVD files
 * Different byte orders in file and host system
 * Probably reading from VTK files that were *not* created by [WriteVTK.jl](https://github.com/jipolanco/WriteVTK.jl) will fail, specifically since
   * compressed data is assumed to be stored as a single block
   * appended data is assumed to be stored as `raw`
   * `header_type` is hardcoded to `UInt64`
+* Extracting primitives from `PolyData` files other than vortices, lines and/or polygons
 * Likely anything else that is not specifically mentioned under *What works*
 
 ## Development
@@ -106,6 +108,8 @@ Its principal developers are
 Further contributions to ReadVTK have been made by the following people:
 * [Jorge Pérez Zerpa](https://www.fing.edu.uy/~jorgepz/)
 (Universidad de la República, Uruguay)
+* [Ondřej Kincl](https://www2.karlin.mff.cuni.cz/~kincl/)
+(Charles University, Czech Republic)
 
 ## License and contributing
 ReadVTK is licensed under the MIT license (see [LICENSE.md](LICENSE.md)).
--- a/README.md
+++ b/README.md
@@ -74,7 +74,7 @@ Further example VTK files can be found in the
 * Extracting point coordinates
 * Extracting information about cell types
 * Only for `ImageData` files: get origin, spacing, and extent information
-* Reading `PolyData` files containing vortices, lines and/or polygons
+* Reading `PolyData` files containing vortices, lines, and/or polygons
 
 ### What does not work
 * Reading VTK files not stored in the VTK XML format
--- a/README.md
+++ b/README.md
@@ -78,7 +78,7 @@ Further example VTK files can be found in the
 
 ### What does not work
 * Reading VTK files not stored in the VTK XML format
-* Reading VTK files of other type than `UnstructuredGrid`, `ImageData` or `PolyData`
+* Reading VTK files of other type than `UnstructuredGrid`, `ImageData`, or `PolyData`
 * Multiblock files, PVD files
 * Different byte orders in file and host system
 * Probably reading from VTK files that were *not* created by [WriteVTK.jl](https://github.com/jipolanco/WriteVTK.jl) will fail, specifically since
--- a/README.md
+++ b/README.md
@@ -85,7 +85,7 @@ Further example VTK files can be found in the
   * compressed data is assumed to be stored as a single block
   * appended data is assumed to be stored as `raw`
   * `header_type` is hardcoded to `UInt64`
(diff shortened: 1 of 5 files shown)


test/runtests.jl
```diff
@@ -257,6 +259,54 @@ mkpath(TEST_EXAMPLES_DIR)
...
+  # Test set for PolyData
+  @testset "PolyData" begin
+    ## Generate sample vtp file
+    
+    # define points
+    n = 20
+    points = zeros(3,n)
+    points[1,:] .= [cos(4*pi*i/n) for i in 0:n-1]
+    points[2,:] .= [sin(4*pi*i/n) for i in 0:n-1]
+    points[3,:] .= [0.2i for i in 0:n-1]
```
ranocha, line 272: Some minor formatting stuff that can help quite a bit when working in larger projects (also below). 
```suggestion
    points = zeros(3, n)
    points[1, :] .= [cos(4 * pi * i / n) for i in 0:(n - 1)]
    points[2, :] .= [sin(4 * pi * i / n) for i in 0:(n - 1)]
    points[3, :] .= [0.2 * i for i in 0:n-1]
```

test/runtests.jl
```diff
@@ -257,6 +259,54 @@ mkpath(TEST_EXAMPLES_DIR)
...
+
+    # define values
+    point_values = [4*pi*i/n for i in 0:n-1]
+    poly_values = [0.3i for i in 1:n-3]
+
+    # save using WriteVTK
+    path = joinpath(TEST_EXAMPLES_DIR, "spiral")
+    vtk_grid(path, points, polys) do vtk
+      vtk["theta", VTKPointData()] = point_values # scalar field attached to points
+      vtk["h", VTKCellData()] = poly_values    # scalar field attached to cells
```
ranocha, line 285:
```suggestion
      vtk["theta", VTKPointData()] = point_values # scalar field attached to points
      vtk["h", VTKCellData()] = poly_values       # scalar field attached to cells
```

test/runtests.jl
```diff
@@ -257,6 +259,54 @@ mkpath(TEST_EXAMPLES_DIR)
...
+    end
+
+    # read data from the vtp file
+    vtk = VTKFile(path*".vtp")
+
+    # check getter functions
+    @testset "get points" begin
+      _points = get_points(vtk)
+      @test all(points .== _points)
```
ranocha, line 294: Do you need
```julia
@test all(points .== _points)
```
or does
```julia
@test points == _points
```
also work? If so, I would prefer the latter version (also two more times below).

src/ReadVTK.jl
```diff
@@ -141,6 +141,10 @@ function VTKFile(filename)
     n_points_per_grid_dir = [whole_extent[2*i]+1 for i in (1:3)]
     n_points = prod(n_points_per_grid_dir)
     n_cells = prod(n_points_per_grid_dir .- 1)
+  elseif file_type == "PolyData"
+    piece = root[file_type][1]["Piece"][1]
+    n_points = parse(Int, attribute(piece, "NumberOfPoints", required=true))
+    n_cells = 0
```
ranocha, line 147: In the tests, you also load cell data. So there must be some way to get the number of cells, isn't there? But as far as I understand, cell data (such as the polygon data you used in the tests) is not required and there could be several kinds of additional data besides the points. Since it works in your case and you would like to use it soon, we could proceed as follows: 
What about adding some comments here about this issue and setting `c_cells= typemin(Int)`? This is obviously a value that must be wrong and doesn't necessarily indicate that there are no cells (such as `n_cells = 0` right now).

ranocha (changes requested):
Thanks a lot for the PR! Please find some comments below.

- [x] Could you please also update the README.md and docs/src/index.md files with the new file type information (in the sections what works, what does not work)? To be on the safe side, you could add that it works with some `PolyData`.
- [x] Could you please add your name to the list of contributors in the same files?

src/ReadVTK.jl
OndrejKincl: I set n_cells = number of vertices + number of lines + number of polygons (which I think makes sense). I have not found a way how to obtain this number directly.

OndrejKincl:
I made several more changes:
* I set n_cells for PolyData to be sum of vertices + lines + polygons
* I realized that cell information cannot be obtained using the `get_cells` function. This is because PolyData can have cells of different kinds and have no attribute "Types". So I made separate struct for them, which I called `VTKPrimitives` and a function `get_primitives` to extract them. In case you don't like it, I can return these changes.
* I added another test for PolyData
* I updated the README

README.md
```diff
@@ -74,16 +74,18 @@ Further example VTK files can be found in the
 * Extracting point coordinates
 * Extracting information about cell types
 * Only for `ImageData` files: get origin, spacing, and extent information
+* Reading `PolyData` files containing vortices, lines and/or polygons
```
ranocha, line 77: Could you please copy these changes also to `docs/src/index.md`?

src/ReadVTK.jl
ranocha: I would have associated the number of cells with higher-dimensional constructs, i.e., not vertices and points when there are polygons. However, I am not a VTK expert, so I don't know their manual by heart. What about this:
- You revert the change to `n_cells` (and `get_integer`, which is only required here, I think?) and set 
  ```julia
  # TODO: decide how to handle the number of cells correctly, see  
  #       https://github.com/trixi-framework/ReadVTK.jl/pull/11
  n_cells = typemin(Int)
  ```
  there.
- Then, I will have a final look, merge this PR soon, and release a new version that you can use for your application motivating you to make this PR.
- You create a new (draft) PR with the changes to `n_cells` here and we will wait until our VTK expert @sloede is available again.

ranocha: Okay, let's keep it this way.

OndrejKincl:
Thank you very much for being a reviewer. I believe it is not easy.

PR #16, opened 2023-02-24
Support RectilinearGrid files

boriskaus:
This adds support for RectilinearGrid (`*.vtr`) files, and includes routines to retrieve 1D coordinate vectors. 
It also includes `get_data_reshaped` which reshapes the structured data in the appropriate formatfor `ImageData` and `RectilinearGrid`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -69,16 +69,17 @@ Further example VTK files can be found in the
 [`ReadVTK_examples` repository](https://github.com/trixi-framework/ReadVTK_examples).
 
 ### What works
-* Reading in VTK XML files of type `UnstructuredGrid` or `ImageData`
+* Reading in VTK XML files of type `UnstructuredGrid`, `RectilinearGrid`,`ImageData`, or `PolyData`
 * Extracting cell or point data
 * Extracting point coordinates
 * Extracting information about cell types
 * Only for `ImageData` files: get origin, spacing, and extent information
+* Only for `RectilinearGrid` files: get 1D coordinate vectors 
 * Reading `PolyData` files containing vortices, lines, and/or polygons
 
 ### What does not work
 * Reading VTK files not stored in the VTK XML format
-* Reading VTK files of other type than `UnstructuredGrid`, `ImageData`, or `PolyData`
+* Reading VTK files of other type than `UnstructuredGrid`, `ImageData`, `RectilinearGrid` or `PolyData`
 * Multiblock files, PVD files
 * Different byte orders in file and host system
 * Probably reading from VTK files that were *not* created by [WriteVTK.jl](https://github.com/jipolanco/WriteVTK.jl) will fail, specifically since
--- a/README.md
+++ b/README.md
@@ -79,7 +79,7 @@ Further example VTK files can be found in the
 
 ### What does not work
 * Reading VTK files not stored in the VTK XML format
-* Reading VTK files of other type than `UnstructuredGrid`, `ImageData`, `RectilinearGrid` or `PolyData`
+* Reading VTK files of other type than what is listed under *What works* above
 * Multiblock files, PVD files
 * Different byte orders in file and host system
 * Probably reading from VTK files that were *not* created by [WriteVTK.jl](https://github.com/jipolanco/WriteVTK.jl) will fail, specifically since
--- a/README.md
+++ b/README.md
@@ -111,6 +111,8 @@ Further contributions to ReadVTK have been made by the following people:
 (Universidad de la República, Uruguay)
 * [Ondřej Kincl](https://www2.karlin.mff.cuni.cz/~kincl/)
 (Charles University, Czech Republic)
+* [Boris Kaus](https://www.geosciences.uni-mainz.de/geophysics-and-geodynamics/team/univ-prof-dr-boris-kaus/)
+  (Johannes-Gutenberg University Mainz, Germany)
 
 ## License and contributing
 ReadVTK is licensed under the MIT license (see [LICENSE.md](LICENSE.md)).
--- a/src/ReadVTK.jl
+++ b/src/ReadVTK.jl
@@ -9,8 +9,9 @@ using LightXML: LightXML, XMLDocument, XMLElement, parse_string, attribute, has_
                 child_elements, free, content, find_element
 
 export VTKFile, VTKData, VTKDataArray, VTKCells, VTKPrimitives,        # structs
-       get_point_data, get_cell_data, get_data,                        # get data functions
+       get_point_data, get_cell_data, get_data, get_data_reshaped,     # get data functions
        get_points, get_cells, get_origin, get_spacing, get_primitives, # get geometry functions
+       get_coordinates, get_coordinate_data,                           # get geometry functions
        get_example_file                                                # other functions
 
 """
@@ -112,7 +113,7 @@ function VTKFile(filename)
   end
 
   # Ensure matching file types
-  if !(file_type in ("UnstructuredGrid", "ImageData", "PolyData"))
+  if !(file_type in ("UnstructuredGrid", "ImageData", "PolyData","RectilinearGrid"))
     error("Unsupported file type: ", file_type)
   end
 
@@ -136,6 +137,12 @@ function VTKFile(filename)
     n_points = parse(Int, attribute(piece, "NumberOfPoints", required=true))
     n_cells = parse(Int, attribute(piece, "NumberOfCells", required=true))
   elseif file_type == "ImageData"
+    dataset_element = root[file_type][1]
+    whole_extent = parse.(Int, split(attribute(dataset_element, "WholeExtent", required=true), ' '))
+    n_points_per_grid_dir = [whole_extent[2*i]+1 for i in (1:3)]
+    n_points = prod(n_points_per_grid_dir)
+    n_cells = prod(n_points_per_grid_dir .- 1)
+  elseif file_type == "RectilinearGrid"
     dataset_element = root[file_type][1]
     whole_extent = parse.(Int, split(attribute(dataset_element, "WholeExtent", required=true), ' '))
     n_points_per_grid_dir = [whole_extent[2*i]+1 for i in (1:3)]
@@ -226,6 +233,15 @@ Retrieve a lightweight `VTKData` object with the point data of the given VTK fil
 See also: [`VTKData`](@ref), [`get_cell_data`](@ref)
 """
 get_point_data(vtk_file) = get_data_section(vtk_file, "PointData")
+
+"""
+  get_coordinate_data(vtk_file)
+
+Retrieve a lightweight `VTKData` object with the coordinates data of the given VTK file.
+
+See also: [`VTKData`](@ref), [`get_point_data`](@ref),  [`get_cell_data`](@ref)
+"""
+get_coordinate_data(vtk_file) = get_data_section(vtk_file, "Coordinates")
 
 
 # Auxiliary methods for conveniently using `VTKData` objects like a dictionary/collectible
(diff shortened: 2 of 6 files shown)


README.md
```diff
@@ -69,16 +69,17 @@ Further example VTK files can be found in the
...
 * Extracting point coordinates
 * Extracting information about cell types
 * Only for `ImageData` files: get origin, spacing, and extent information
+* Only for `RectilinearGrid` files: get 1D coordinate vectors 
 * Reading `PolyData` files containing vortices, lines, and/or polygons
 
 ### What does not work
 * Reading VTK files not stored in the VTK XML format
-* Reading VTK files of other type than `UnstructuredGrid`, `ImageData`, or `PolyData`
+* Reading VTK files of other type than `UnstructuredGrid`, `ImageData`, `RectilinearGrid` or `PolyData`
```
sloede, line 82:
```suggestion
* Reading VTK files of other type than what is listed under *What works* above
```
I think this point was ok while there were one or two supported file formats; thanks to your contribution we finally have too many to list again here :-)

src/ReadVTK.jl
```diff
@@ -112,7 +113,7 @@ function VTKFile(filename)
   end
 
   # Ensure matching file types
-  if !(file_type in ("UnstructuredGrid", "ImageData", "PolyData"))
+  if !(file_type in ("UnstructuredGrid", "ImageData", "PolyData","RectilinearGrid"))
```
sloede, line 116:
```suggestion
  if !(file_type in ("UnstructuredGrid", "ImageData", "PolyData", "RectilinearGrid"))
```

src/ReadVTK.jl
```diff
@@ -141,6 +142,12 @@ function VTKFile(filename)
     n_points_per_grid_dir = [whole_extent[2*i]+1 for i in (1:3)]
     n_points = prod(n_points_per_grid_dir)
     n_cells = prod(n_points_per_grid_dir .- 1)
+  elseif file_type == "RectilinearGrid"
+    dataset_element = root[file_type][1]
+    whole_extent = parse.(Int, split(attribute(dataset_element, "WholeExtent", required=true), ' '))
+    n_points_per_grid_dir = [whole_extent[2*i]+1 for i in (1:3)]
+    n_points = prod(n_points_per_grid_dir)
+    n_cells = prod(n_points_per_grid_dir .- 1)
```
sloede, line 150: If I am not mistaken, this is the exact same code as for files of type `ImageData`. Maybe just modify the if condition in line 139 above to
```julia
elseif file_type == "ImageData" || file_type == "RectilinearGrid"
```
?

src/ReadVTK.jl
```diff
@@ -227,6 +234,15 @@ See also: [`VTKData`](@ref), [`get_cell_data`](@ref)
 """
 get_point_data(vtk_file) = get_data_section(vtk_file, "PointData")
 
+"""
+  get_coordinate_data(vtk_file)
+
+Retrieve a lightweight `VTKData` object with the coordinates data of the given VTK file.
```
sloede, line 240:
```suggestion
    get_coordinate_data(vtk_file)

Retrieve a lightweight `VTKData` object with the coordinate data of the given VTK file.
```

src/ReadVTK.jl
```diff
@@ -443,6 +459,54 @@ function get_data(data_array::VTKDataArray{T,N,<:FormatAppended}) where {T,N}
...
+
+  return data_reshape
+ 
+end
+
+"""
+  get\\_local\\_size(xml_file, file_type, cell_data=false)
+
+Retrieve the local size of a structured grid (ImageData, RectilinearGrid). Note that this always returns 3 dimensions, even if the data is 1D or 2D. 
+"""
```
sloede, line 492:
```suggestion
"""
    get_local_size(xml_file, file_type, cell_data=false)

Retrieve the local size of a structured grid (ImageData, RectilinearGrid). Note that this always returns three dimensions, even if the data is 1D or 2D. 
"""
```

src/ReadVTK.jl
```diff
@@ -443,6 +459,54 @@ function get_data(data_array::VTKDataArray{T,N,<:FormatAppended}) where {T,N}
...
+function get_local_size(xml_file, file_type, cell_data=false)
+
```
sloede, line 494:
```suggestion
```
Style convention: no blank line at the beginning of function body

src/ReadVTK.jl
```diff
@@ -443,6 +459,54 @@ function get_data(data_array::VTKDataArray{T,N,<:FormatAppended}) where {T,N}
...
+  return data_reshape
+ 
+end
+
+"""
+  get\\_local\\_size(xml_file, file_type, cell_data=false)
+
+Retrieve the local size of a structured grid (ImageData, RectilinearGrid). Note that this always returns 3 dimensions, even if the data is 1D or 2D. 
+"""
+function get_local_size(xml_file, file_type, cell_data=false)
```
sloede, line 493: Why is this called "local" size? Is there another one or why is it specifically a local size?

src/ReadVTK.jl
```diff
@@ -443,6 +459,54 @@ function get_data(data_array::VTKDataArray{T,N,<:FormatAppended}) where {T,N}
...
+
+  N = whole_extent[2:2:end] - whole_extent[1:2:end-1] .+ 1
+
+  if cell_data
+    N = N .- 1
+    N[N.==0] .= 1
+  end
+
+
+  return N
```
sloede, line 507:
```suggestion
  Nlocal_size = whole_extent[2:2:end] - whole_extent[1:2:end-1] .+ 1

  if cell_data
    local_size = local_size .- 1
    local_size[local_size .== 0] .= 1
  end

  return local_size
```
We often use `N` as a type parameter, thus I'd suggest a more descriptive name. Also, we use capital names only for types or type parameters, not for variables.

src/ReadVTK.jl
```diff
@@ -468,6 +532,31 @@ function get_points(vtk_file)
...
+  get_coordinates(vtk_file; x_string="x",y_string="y",z_string="z")
+
+Retrieve VTK coordinate vectors in each direction as 1D vectors for a RectilinearGrid file
+
+The points are given as 1D vectors. Note that in VTK, points are always stored
+three-dimensional, even for 1D or 2D files, so you will always retrieve 3 vectors.
+The 
+
+See also: [`get_cells`](@ref)
+"""
```
sloede, line 545:
```suggestion
"""
    get_coordinates(vtk_file; x_string="x", y_string="y", z_string="z")

Retrieve VTK coordinate vectors in each direction as a tuple of 1D vectors for a RectilinearGrid file.

Note that in VTK, points are always stored three-dimensional, even for 1D or 2D files, so you will always retrieve a tuple with three vectors.

See also: [`get_cells`](@ref)
"""
```

src/ReadVTK.jl
```diff
@@ -468,6 +532,31 @@ function get_points(vtk_file)
...
+function get_coordinates(vtk_file; x_string="x",y_string="y",z_string="z")
+
+  if vtk_file.file_type != "RectilinearGrid"
+      error("the file_type must be RectilinearGrid.")
+  end
```
sloede, line 550:
```suggestion
function get_coordinates(vtk_file; x_string="x", y_string="y", z_string="z")
  if vtk_file.file_type != "RectilinearGrid"
      error("The file type of the VTK file must be 'RectilinearGrid' (current: $(vtk_file.file_type)).")
  end
```

src/ReadVTK.jl
```diff
@@ -468,6 +532,31 @@ function get_points(vtk_file)
...
+
+  coordinates = get_coordinate_data(vtk_file)
+  x = get_data(coordinates[x_string])
+  y = get_data(coordinates[y_string])
+  z = get_data(coordinates[z_string])
+
+  return  x,y,z
```
sloede, line 557:
```suggestion
  return  x, y, z
```

src/ReadVTK.jl
```diff
@@ -443,6 +459,54 @@ function get_data(data_array::VTKDataArray{T,N,<:FormatAppended}) where {T,N}
...
 
 
+"""
+    get\\_data\\_reshaped(data_array::VTKDataArray; cell_data=false)
+
+Retrieve actual data from a `VTKDataArray` and reshapes them as 1D, 2D or 3D arrays, in case we deal with structured grids.
+Note that vectors or tensors will have their components stored in the first dimension of the array. 
+As there is no way to automatically tell from the VTK file format whether it is a tensor, the user has to reshape this accordingly.
+
+"""
```
sloede, line 469:
```suggestion
"""
    get_data_reshaped(data_array::VTKDataArray; cell_data=false)

Retrieve actual data from a `VTKDataArray` and reshape it as 1D, 2D, or 3D arrays, in case we deal with structured grids.
Note that vectors or tensors will have their components stored in the first dimension of the array. 
As there is no way to automatically tell from the VTK file format whether it is a tensor, the user has to reshape this manually.
"""
```
We don't need to escape the underscores here.

src/ReadVTK.jl
```diff
@@ -443,6 +459,54 @@ function get_data(data_array::VTKDataArray{T,N,<:FormatAppended}) where {T,N}
...
+function get_data_reshaped(data_array::VTKDataArray{T,Ncomponents}; cell_data=false) where {T,Ncomponents}
```
sloede, line 470:
```suggestion
function get_data_reshaped(data_array::VTKDataArray{T,N}; cell_data=false) where {T,N}
```
In the actual type we also use `N` for this type parameter, thus I'd prefer to stay consistent here.

src/ReadVTK.jl
```diff
@@ -443,6 +459,54 @@ function get_data(data_array::VTKDataArray{T,N,<:FormatAppended}) where {T,N}
...
+
```
sloede, line 471:
```suggestion
```

src/ReadVTK.jl
```diff
@@ -443,6 +459,54 @@ function get_data(data_array::VTKDataArray{T,N,<:FormatAppended}) where {T,N}
...
+  # reshape 
+  if Ncomponents==1
+    data_reshape = reshape(data, N...)
+  else
+    data_reshape = reshape(data, Ncomponents, N...)
+  end
+
+  return data_reshape
+ 
+end
```
sloede, line 486:
```suggestion
  # Retrieve size of grid
  local_size = get_local_size(data_array.vtk_file.xml_file, data_array.vtk_file.file_type, cell_data)
  
  # reshape 
  if N == 1
    data_reshaped = reshape(data, local_size...)
  else
    data_reshaped = reshape(data, N, local_size...)
  end

  return data_reshaped
end
```
As above, use `N` for the parameter and `local_size` for the variable name.

test/runtests.jl
```diff
@@ -212,16 +212,19 @@ mkpath(TEST_EXAMPLES_DIR)
     @testset "get scalar cell data" begin
       cell_data_raw = get_data(get_cell_data(vtk)[cell_data_name])
       cell_data_reshaped = reshape(cell_data_raw, ((Nx-1), (Ny-1), (Nz-1)))
-      @test cell_data_reshaped == cell_scalar_field
+      cell_data_reshaped1 = get_data_reshaped(get_cell_data(vtk)[cell_data_name], cell_data=true)
+      
+      @test cell_data_reshaped == cell_scalar_field
```
sloede, line 217:
```suggestion
      @test cell_data_reshaped == cell_scalar_field
      @test cell_data_reshaped1 == cell_scalar_field
```
I think you forgot to add a test for your newly retrieved data, or am I missing something?

test/runtests.jl
```diff
@@ -212,16 +212,19 @@ mkpath(TEST_EXAMPLES_DIR)
...
+      @test cell_data_reshaped == cell_scalar_field 
     end
 
     @testset "get scalar point data" begin
       point_data_raw = get_data(get_point_data(vtk)[point_data_name])
       point_data_reshaped = reshape(point_data_raw, (Nx, Ny, Nz))
-      @test point_data_reshaped == point_scalar_field
+      point_data_reshaped1 =  get_data_reshaped(get_point_data(vtk)[point_data_name])
+      
+      @test point_data_reshaped == point_scalar_field == point_data_reshaped1
```
sloede, line 225:
```suggestion
      @test point_data_reshaped == point_scalar_field
      @test point_data_reshaped1 == point_scalar_field
```
I'd make two tests out of it such that we get the line number in case only one of them fails

test/rectilinear.jl
```diff
@@ -0,0 +1,138 @@
+using StaticArrays: SVector, SMatrix
+
+const FloatType = Float32
+const vtk_filename_noext = joinpath(TEST_EXAMPLES_DIR,"rectilinear")
```
sloede, line 4:
```suggestion
const vtk_filename_noext = joinpath(TEST_EXAMPLES_DIR, "rectilinear")
```

test/rectilinear.jl
```diff
@@ -0,0 +1,138 @@
...
+
+outfiles = String[]
+
+for compress in [true, false]
+for dim in 2:3
```
sloede, line 9: Could you please indent this block as well? Otherwise I will always stumble over this whenever I read it :-) 

Also, we typically use 2 spaces instead of 4 spaces for indentation. It would be great if you could adapt this in this file for consistency with the other files in the package.

sloede (changes requested):
Thanks a lot for your contribution! Overall it looks to be in great shape already, I only left a few minor comments/suggestions.

Please also add your name to the author's list in the [README.md](https://github.com/trixi-framework/ReadVTK.jl#authors).

test/rectilinear.jl
boriskaus: fine

boriskaus:
thanks a lot for the quick review! I've added the requested changes.

src/ReadVTK.jl
```diff
@@ -59,6 +60,7 @@ end
 # Header type is hardcoded and corresponds to VTK XML version 1.0
 header_type(::VTKFile) = UInt64
 
+
```
sloede, line 63:
```suggestion
```

src/ReadVTK.jl
```diff
@@ -125,7 +126,7 @@ function VTKFile(filename)
            (byte_order == "BigEndian" && !is_little_endian))
 
   # Ensure matching header type
-  @assert header_type == "UInt64"
+  #@assert header_type == "UInt64"
```
sloede, line 129: Why was this commented? Did I miss. This in the first round?

boriskaus: sorry, leftover of local debugging. We are using an older file format in our [code](https://bitbucket.org/bkaus/lamem/src/master/) and were trying to see if we can read those files with the package - no.

src/ReadVTK.jl
boriskaus: if we have parallel `*.pvtr` files, each of the `*.vtr` files will only store the local portion. This is not yet supported, but may be supported in a future version.

sloede: Thanks for the clarification! /r

sloede (approved):
LGTM! Thanks again for the contribution @boriskaus!

PR #17, opened 2023-03-01
Support parallel PVTK and PVD files

boriskaus:
This adds support for `PVD` and `PVTK` files, specifically for `*.pvti`,`*.pvtu` and `*.ptr` files, which allows loading the corresponding parallel file. In the case of structured grids (`ImageData` or `RectiLinearGrid`) the individual pieces of the grid are collected into the full grid.


src/get_functions.jl
```diff
@@ -23,13 +23,22 @@ end
 
 
 """
-    get_spacing(vtk_file)
+    get_origin(vtk_file::PVTKFile)
+
+Retrieve the vector of coordinates of the origin of a uniform grid from the given [`PVTKFile`](@ref) file.
+
+See also: [`VTKFile`](@ref)
```
sloede, line 30:
```suggestion
See also: [`PVTKFile`](@ref)
```

src/get_functions.jl
```diff
@@ -61,15 +79,24 @@ function get_whole_extent(vtk_file)
   return whole_extent
 end
```
sloede, line 80: I cannot attach this comment to the correct line number, but for all other functions you added a type specialization to the function argument as `vtk_file::VTKFile` for existing methods. For consistency, I think you should also add this to the existing `get_whole_extent` method.

test/runtests.jl
```diff
@@ -366,6 +366,10 @@ mkpath(TEST_EXAMPLES_DIR)
     include("rectilinear.jl")
   end
 
+  @testset "pvtk files" begin
```
sloede, line 369:
```suggestion
  @testset "Parallel VTK (PVTK) files" begin
```
Just a little more information for users who are not as familiar with the parallel file formats.

sloede (changes requested):
Hi Boris! Thanks again for this great contribution. I am currently on a conference and do not have much time, but most of the implementation looks quite good. This is only the first part of the review though, I still haven't looked at `src/ReadVTK.jl` or `test/pvtk_files.jl`, but hope to do so in the next two days.

src/ReadVTK.jl
```diff
@@ -172,6 +172,169 @@ end
 # Return `Piece` XML element that contains all VTK data
 piece(vtk_file::VTKFile) = LightXML.root(vtk_file.xml_file)[vtk_file.file_type][1]["Piece"][1]
 
+"""
+  isstructured(xml_file::XMLDocument)
+Returns `true` is it is a structured grid
+"""
+function isstructured(xml_file::XMLDocument)
```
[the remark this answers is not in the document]
boriskaus: ok, didn't know about the 4 spaces. I generally thought that type annotations are better for example to speed up code compilation. In this case the routine won't work if `xml_file` is not an `XMLDocument`, so we might as well add it, not?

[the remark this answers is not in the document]
sloede: Type annotations are really *only* needed to make different *methods* (i.e., "functions with the same name") distinguishable for the compiler. That is, you have a `get_origin(file)` implementation for `file` of type `VTKFile` and a different one for `file` of type `PVTKFile`. In that case, the compiler needs the type annotations `file::VTKFile` etc. to know which version to pick.

Adding a type annotation will _never_ have an impact on performance (at least in all cases that I am aware of).

> In this case the routine won't work if `xml_file` is not an `XMLDocument`, so we might as well add it, not?

We had this discussion in Trixi a few times as well. In many cases it is not clear whether the same implementation could also be used by another type successfully. Therefore, we decided that we will never add type annotations except to make it possible for the compiler to dispatch on the type for different implementations.

src/get_functions.jl
```diff
@@ -43,6 +52,15 @@ function get_spacing(vtk_file)
...
+
+Retrieve a vector with the regular increments in each coordinate direction of the uniform grid from the given [`PVTKFile`](@ref) file.
+
+See also: [`PVTKFile`](@ref)
+"""
+get_spacing(pvtk_file::PVTKFile) = get_spacing(VTKFile(pvtk_file.saved_files[1]))
+
 
 """
     get_whole_extent(vtk_file)
```
sloede, line 66:
```suggestion
    get_whole_extent(vtk_file::VTKFile)
```
To be consistent with the code.

test/pvtk_files.jl
```diff
@@ -0,0 +1,187 @@
...
+    for (n, time) ∈ enumerate(times)
+        vtk_grid("timestep_$n", x, y, z) do vtk
+            vtk["Pressure"] = rand(length(x), length(y), length(z))
+            pvd[time] = vtk
+        end
+    end
+end
+
+
+# (2) Read back files
```
sloede, line 106: I would really recommend to structure the tests into further `@testset`s. At the very least, I would create one testset for rectilinear, one for image data, one for PVTU, and one for PVD files. The motivation is that this makes it easier in the output of the test runs to figure out what has failed, as it gives you not just the line number but a name attached to it. You could even add more than one layer of `@testset`s, but at least the main parts I would separate into separate testsets

test/pvtk_files.jl
```diff
@@ -0,0 +1,187 @@
...
+
+# write *.pvtr file
+for part = 1:4
+  is, js, ks = extents[part]  # local indices
+  xs, ys, zs = xs_global[is], ys_global[js], zs_global[ks]  # local grid
+  xs_c, ys_c, zs_c = xs[1:end-1], ys[1:end-1], zs[1:end-1]
+  saved_files[part] = pvtk_grid(
+          "fields", Vector(xs), Vector(ys), Vector(zs);
+          part = part, extents = extents,
+      ) do pvtk
```
sloede, line 46: I attach this here but it really applies to multiple locations:

I strongly recommend to remove all files that are generated by your tests in the beginning, and then again at the end. This way, you minimize the chance of accidentally ending up with an old test file that was generated during a previous run and which could compromise the validity of subsequent test runs.

For this purpose, we create a test file directory in
https://github.com/trixi-framework/ReadVTK.jl/blob/3abe6b8b87bf1e79ec8f34de2158bc20fa5331f2/test/runtests.jl#L19-L21
and delete it at the end again
https://github.com/trixi-framework/ReadVTK.jl/blob/3abe6b8b87bf1e79ec8f34de2158bc20fa5331f2/test/runtests.jl#L373.

If you use a path to the test directory here and in the other locations, you will automatically have everything removed in the end again:
```suggestion
  path = joinpath(TEST_EXAMPLES_DIR, "fields")
  saved_files[part] = pvtk_grid(
          path, Vector(xs), Vector(ys), Vector(zs);
          part = part, extents = extents,
      ) do pvtk
```
See also e.g. here:
https://github.com/trixi-framework/ReadVTK.jl/blob/3abe6b8b87bf1e79ec8f34de2158bc20fa5331f2/test/runtests.jl#L347-L350

sloede (changes requested):
Second part of my review, addressing mostly the testing setup. `src/ReadVTK.jl` still missing (but hopefully coming soon)

test/pvtk_files.jl
boriskaus: yes, very good point - thanks!

test/pvtk_files.jl
boriskaus: good point; I've added that

src/ReadVTK.jl
```diff
@@ -475,28 +758,113 @@ function get_data_reshaped(data_array::VTKDataArray{T,N}; cell_data=false) where
...
 end
 
 """
-    get_points(vtk_file)
+  get_extents(xml_file, min_extent=[0;0;0])
+
+Retrieve the local size of pieces of a structured grid (ImageData, RectilinearGrid). Note that this always returns three dimensions, even if the data is 1D or 2D. 
+"""
+function get_extents(xml_file, min_extent=[0;0;0])
+
```
sloede, line 837:
```suggestion
```

src/ReadVTK.jl
```diff
@@ -543,6 +919,30 @@ function get_coordinates(vtk_file; x_string="x", y_string="y", z_string="z")
 end
 
 
+"""
+  get_coordinates(pvtk_file::{VTKFile; x_string="x", y_string="y", z_string="z")
```
sloede, line 923:
```suggestion
    get_coordinates(pvtk_file::{VTKFile; x_string="x", y_string="y", z_string="z")
```

src/ReadVTK.jl
```diff
@@ -543,6 +919,30 @@ function get_coordinates(vtk_file; x_string="x", y_string="y", z_string="z")
...
+
+Retrieve VTK coordinate vectors in each direction as a tuple of 1D vectors for a PRectilinearGrid file.
+
+Note that in VTK, points are always stored three-dimensional, even for 1D or 2D files, so you will always retrieve a tuple with three vectors.
+
+See also: [`get_cells`](@ref)
+"""
+function get_coordinates(pvtk_file::PVTKFile{N}; x_string="x", y_string="y", z_string="z") where N
+
```
sloede, line 932:
```suggestion
```

sloede (changes requested):
I finally got to the last part of the review. There were many many changes in this PR, thus there are quite a few comments 😬 Please consider most of them as suggestions (unless indicated otherwise); they are mostly to make the style and variable names consistent with the rest of the code.

In case of questions, please let me know!

In any case, great work!

src/ReadVTK.jl
```diff
@@ -172,6 +172,173 @@ end
...
+# Fields
+- `filename`: original path to the PVTK file that has been read in
+- `xml_file`: xml info
+- `file_type`: currently only `"PRectilinearGrid"` or `"PImageData"` are supported
+- `version`: currently only v1.0 is supported
+- `saved_files`: vector with strings that contain the filenames of each of the parallel files
+- `vtk`: vector with `VTKFile` data that contains the info about each of the files
+"""
+struct PVTKFile{N}
```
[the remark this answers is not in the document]
boriskaus: ok, I've changed that

src/ReadVTK.jl
```diff
@@ -210,33 +394,73 @@ end
...
+Retrieve a lightweight vector with `PVTKData` objects with the point data of the given PVTK files.
+
+See also: [`PVTKData`](@ref), [`get_cell_data`](@ref)
 """
-    get_coordinate_data(vtk_file)
+function get_point_data(pvtk_file::PVTKFile{N}) where N 
+  pdata_v = VTKData[]
+  for i=1:N
+    push!(pdata_v, get_point_data(pvtk_file.vtk[i]))
+  end
```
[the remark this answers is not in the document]
boriskaus: ok fine. didn't worked initially but does now

src/ReadVTK.jl
```diff
@@ -190,8 +357,25 @@ end
...
+    PVTKData
+
+Convenience type to hold information about available data arrays for cells or points.
+
+Supports a collectible/dictionary-like syntax, e.g., `keys(pvtk_data)` to show available data arrays
+or `pvtk_data["varname"]` to retrieve the `VTKDataArray` for variable `varname`.
+"""
+struct PVTKData
+  data::Vector{VTKData}
+  parent_xml::XMLDocument
```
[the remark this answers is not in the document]
boriskaus: agree

src/ReadVTK.jl
```diff
@@ -451,6 +722,18 @@ function get_data(data_array::VTKDataArray{T,N,<:FormatAppended}) where {T,N}
   end
 end
 
+"""
+    get_data(data_array::PVTKDataArray)
+
+Retrieve actual data from a `PVTKDataArray` as a one- or two-dimensional array-like container.
+"""
+function get_data(data_array::PVTKDataArray) 
+  dat = []
```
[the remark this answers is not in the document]
boriskaus: ok, done now

src/ReadVTK.jl
```diff
@@ -172,6 +172,173 @@ end
...
+
+  # Ensure correct version
+  @assert version == v"1.0"
+
+  # Extract names of files & load the data
+  pieces = root[file_type][1]["DataSet"]
+  N        = length(pieces)
+  file     = Vector{String}(undef, N)  
+  dir      = Vector{String}(undef, N)  
+  timestep = Vector{Float64}(undef,N)
```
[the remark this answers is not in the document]
boriskaus: changed

boriskaus:
Unless I missed some, I believe that all suggestions are now implemented - thanks a lot for that!

src/ReadVTK.jl
```diff
@@ -172,6 +172,173 @@ end
...
+  N      = length(pieces)
+  vtk_filenames = Vector{String}(undef, N)  
+  vtk = Vector{VTKFile}(undef, N)  
+  
+  for i=1:N
+    vtk_filenames[i] = attribute(pieces[i],  "Source", required=true)
+    vtk[i] = VTKFile(joinpath(dir,vtk_filenames[i]))
+  end
+
+  return PVTKFile(filename, xml_file, file_type, version, vtk_filenames, vtk)
```
sloede, line 260:
```suggestion
  # Extract names of files & load the data
  pieces   = root[file_type][1]["Piece"]
  n_pieces = length(pieces)
  vtk_filenames = Vector{String}(undef, n_pieces)  
  vtk_files = Vector{VTKFile}(undef, n_pieces)  
  
  for i in 1:n_pieces
    vtk_filenames[i] = attribute(pieces[i], "Source", required=true)
    vtk_files[i] = VTKFile(joinpath(dir, vtk_filenames[i]))
  end

  return PVTKFile(filename, xml_file, file_type, version, vtk_filenames, vtk_files)
```
For consistency

src/ReadVTK.jl
```diff
@@ -172,6 +172,173 @@ end
...
+  timestep = Vector{Float64}(undef,n_pieces)
+  
+  for i=1:n_pieces
+    file_dir = attribute(pieces[i],  "file", required=true)
+    file[i] = file_dir;
+    dir[i] = dirname(file_dir);
+    timestep[i] = parse(Float64,attribute(pieces[i],  "timestep", required=true))
+  end
+
+  return PVDFile(filename, file_type, file, dir, timestep)
```
sloede, line 328:
```suggestion
  # Extract names of files & load the data
  pieces = root[file_type][1]["DataSet"]
  n_pieces = length(pieces)
  vtk_filenames = Vector{String}(undef, n_pieces)  
  directories = Vector{String}(undef, n_pieces)  
  timesteps = Vector{Float64}(undef, n_pieces)
  
  for i in 1:n_pieces
    file_dir = attribute(pieces[i], "file", required=true)
    vtk_filenames[i] = file_dir
    directories[i] = dirname(file_dir)
    timesteps[i] = parse(Float64, attribute(pieces[i], "timestep", required=true))
  end

  return PVDFile(filename, file_type, vtk_filenames, directories, timesteps)
```
For consistency

src/ReadVTK.jl
```diff
@@ -210,33 +394,75 @@ end
...
 
+Retrieve a lightweight vector with `PVTKData` objects with the cell data of the given PVTK files.
 
+See also: [`PVTKData`](@ref), [`get_cell_data`](@ref)
 """
-    get_point_data(vtk_file)
+function get_cell_data(pvtk_file::PVTKFile)
+  N = length(pvtk_file)
+  cdata_v = Vector{VTKData}(undef,N)
+  for i=1:N
```
sloede, line 415:
```suggestion
  n_files = length(pvtk_file)
  cdata_v = Vector{VTKData}(undef, n_files)
  for i in 1:n_files
```
For consistency

src/ReadVTK.jl
```diff
@@ -210,33 +394,75 @@ end
...
+    get_point_data(pvtk_file::PVTKFile)
+
+Retrieve a lightweight vector with `PVTKData` objects with the point data of the given PVTK files.
+
+See also: [`PVTKData`](@ref), [`get_cell_data`](@ref)
+"""
+function get_point_data(pvtk_file::PVTKFile)
+  N = length(pvtk_file)
+  pdata_v = Vector{VTKData}(undef, N)
+  for i in 1:N
```
sloede, line 442:
```suggestion
  n_files = length(pvtk_file)
  pdata_v = Vector{VTKData}(undef, n_files)
  for i in 1:n_files
```
For consistency

src/ReadVTK.jl
```diff
@@ -263,6 +495,24 @@ function Base.getindex(data::VTKData, name)
...
+function Base.getindex(data::PVTKData, name)
+  index = findfirst(isequal(name), data.data[1].names)
+
+  if isnothing(index)
+    throw(KeyError(name))
+  end
+
+  N = length(data.data)
+  data_array = Vector{VTKDataArray}(undef,N)
+  for i=1:N
```
sloede, line 508:
```suggestion
  n_datasets = length(data.data)
  data_array = Vector{VTKDataArray}(undef, n_datasets)
  for i in 1:n_datasets
```
For consistency

src/ReadVTK.jl
```diff
@@ -451,6 +721,20 @@ function get_data(data_array::VTKDataArray{T,N,<:FormatAppended}) where {T,N}
...
+"""
+    get_data(data_array::PVTKDataArray)
+
+Retrieve actual data from a `PVTKDataArray` as a one- or two-dimensional array-like container.
+"""
+function get_data(data_array::PVTKDataArray) 
+
+  N = length(data_array.data)
+  dat = Vector{Array}(undef, N)  
+  for i in 1:N
```
sloede, line 733:
```suggestion
  n_datasets = length(data_array.data)
  dat = Vector{Array}(undef, n_datasets)  
  for i in 1:n_datasets
```
For consistency

src/ReadVTK.jl
```diff
@@ -451,6 +721,20 @@ function get_data(data_array::VTKDataArray{T,N,<:FormatAppended}) where {T,N}
...
+    dat[i] = get_data(data_array.data[i])
+  end
+  return dat
```
sloede, line 736:
```suggestion
  end
  
  return dat
```
For consistency

src/ReadVTK.jl
```diff
@@ -543,6 +918,30 @@ function get_coordinates(vtk_file; x_string="x", y_string="y", z_string="z")
...
+Note that in VTK, points are always stored three-dimensional, even for 1D or 2D files, so you will always retrieve a tuple with three vectors.
+
+See also: [`get_cells`](@ref)
+"""
+function get_coordinates(pvtk_file::PVTKFile; x_string="x", y_string="y", z_string="z")
+  N = length(pvtk_file)
+  coords =  get_coordinates.(pvtk_file.vtk_files,x_string=x_string, y_string=y_string, z_string=z_string);
+
+  x,y,z = coords[1][1][:],coords[1][2][:],coords[1][3][:]
+  for i=2:N
```
sloede, line 935:
```suggestion
function get_coordinates(pvtk_file::PVTKFile; x_string="x", y_string="y", z_string="z")
  coords =  get_coordinates.(pvtk_file.vtk_files,x_string=x_string, y_string=y_string, z_string=z_string);

  x,y,z = coords[1][1][:],coords[1][2][:],coords[1][3][:]
  for i=2:length(pvtk_file)
```
Simplify

sloede (approved):
LGTM! Thanks a lot for the great contribution @boriskaus! Once all tests pass and if the code coverage numbers look ok, this can be merged.

PR #18, opened 2023-03-06
Support for StructuredGrid files

boriskaus:
This PR adds support for `StructuredGrid` (`*.vts`) files as well as their parallel counterparts `PStructuredGrid` (`*.pvts`)

### DIFF
--- a/src/ReadVTK.jl
+++ b/src/ReadVTK.jl
@@ -174,8 +174,8 @@ end
 piece(vtk_file::VTKFile) = LightXML.root(vtk_file.xml_file)[vtk_file.file_type][1]["Piece"][1]
 
 """
-  isstructured(type::String)
-Returns `true` is it is a struct
+  isstructured(xml_file::XMLDocument)
+Returns `true` is it is a structured grid
 """
 function isstructured(xml_file::XMLDocument)
   root = LightXML.root(xml_file)
--- a/src/ReadVTK.jl
+++ b/src/ReadVTK.jl
@@ -180,10 +180,9 @@ Returns `true` if it is a structured grid.
 function isstructured(xml_file)
   root = LightXML.root(xml_file)
   type = attribute(root, "type", required=true)
-  if type=="RectilinearGrid" || type=="ImageData" || 
-    type=="PRectilinearGrid" || type=="PImageData"
+  if (type == "RectilinearGrid" || type == "ImageData" ||
+      type == "PRectilinearGrid" || type == "PImageData")
     structured = true
-
   else
     structured = false
   end
--- a/src/ReadVTK.jl
+++ b/src/ReadVTK.jl
@@ -700,6 +700,7 @@ function get_data(data_array::VTKDataArray{T,N,<:FormatAppended}) where {T,N}
 
     # Pass data through ZLib decompressor
     if last>length(raw)
+      @show data_array data_array.vtk_file.xml_file
       @show first, last, size(raw)
       error("mistake in get_data")
     end
--- a/src/ReadVTK.jl
+++ b/src/ReadVTK.jl
@@ -205,20 +205,20 @@ Hold all relevant information about a Parallel VTK XML file that has been read i
 - `vtk`: vector with `VTKFile` data that contains the info about each of the files
 """
 struct PVTKFile{N}
-  filename::String
+  filename::String      # filename
   xml_file::XMLDocument
   file_type::String
   version::VersionNumber
-  saved_files::Vector{String}                  # filenames
+  saved_files::Vector{String}                # filenames
   vtk::Vector{VTKFile}
 end
 
 """
-    PVTKFile(filename)
+    PVTKFile(filename; dir="")
 
-Read in and parse the PVTK XML file specified by its `filename`.
+Read in and parse the PVTK XML file specified by its `filename`. Optionally, an additional directory name `dir` can be specified.
 """
-function PVTKFile(filename)
+function PVTKFile(filename; dir="")
   # Read in file into memory as a string
   raw_file_contents = read(filename, String)
 
@@ -253,7 +253,7 @@ function PVTKFile(filename)
   
   for i=1:N
     saved_files[i] = attribute(pieces[i],  "Source", required=true)
-    vtk[i] = VTKFile(saved_files[i])
+    vtk[i] = VTKFile(joinpath(dir,saved_files[i]))
   end
 
   return PVTKFile{N}(filename, xml_file, file_type, version, saved_files, vtk)
@@ -279,6 +279,7 @@ struct PVDFile{N}
   filename::String
   file_type::String
   file::Vector{String}            # filenames
+  dir::Vector{String}             # directory names
   timestep::Vector{Float64}       # times
 end
 
@@ -315,14 +316,17 @@ function PVDFile(filename)
   pieces = root[file_type][1]["DataSet"]
   N        = length(pieces)
   file     = Vector{String}(undef, N)  
+  dir      = Vector{String}(undef, N)  
   timestep = Vector{Float64}(undef,N)
   
   for i=1:N
-    file[i] = attribute(pieces[i],  "file", required=true)
+    file_dir = attribute(pieces[i],  "file", required=true)
+    file[i] = file_dir;
+    dir[i] = dirname(file_dir);
     timestep[i] = parse(Float64,attribute(pieces[i],  "timestep", required=true))
   end
 
-  return PVDFile{N}(filename, file_type, file, timestep)
+  return PVDFile{N}(filename, file_type, file, dir, timestep)
 end
 
 # Reduce noise:
@@ -737,8 +741,8 @@ function get_data_reshaped(data_array::VTKDataArray{T,N}; cell_data=false) where
   data = get_data(data_array);
 
   # Retrieve size of grid
-  local_size = get_wholeextent(data_array.vtk_file.xml_file, cell_data)
-
+  local_size, _ = get_wholeextent(data_array.vtk_file.xml_file, cell_data)
+  
   # reshape 
   if N == 1
     data_reshaped = reshape(data, local_size...)
@@ -757,8 +761,9 @@ It also puts it in the correct location the the full grid
 """
 function get_data_reshaped(data_array::PVTKDataArray; cell_data=false)
   
-  extents = get_extents( data_array.parent_xml)         # local extents
-  wholeextent = get_wholeextent(data_array.parent_xml)  # global grid size
+  wholeextent, min_extent = get_wholeextent(data_array.parent_xml)  # global grid size
+  extents = get_extents( data_array.parent_xml, min_extent)         # local extents
+
   if cell_data
     wholeextent = wholeextent .- 1
   end
@@ -805,22 +810,25 @@ function get_wholeextent(xml_file, cell_data=false)
   dataset_element = root[file_type][1]
   whole_extent = parse.(Int, split(attribute(dataset_element, "WholeExtent", required=true), ' '))
 
-  local_size = whole_extent[2:2:end] - whole_extent[1:2:end-1] .+ 1
+  min_ex = whole_extent[1:2:end-1]
+  max_ex = whole_extent[2:2:end]
+  
+  local_size = [length(min_ex[i]:max_ex[i]) for i=1:3]
 
   if cell_data
     local_size = local_size .- 1
     local_size[local_size .== 0] .= 1
   end
 
-  return local_size
+  return local_size, min_ex
 end
 
 """
-  get_extents(xml_file)
+  get_extents(xml_file, min_extent=[0;0;0])
 
 Retrieve the local size of pieces of a structured grid (ImageData, RectilinearGrid). Note that this always returns three dimensions, even if the data is 1D or 2D. 
 """
-function get_extents(xml_file)
+function get_extents(xml_file, min_extent=[0;0;0])
 
   if !isstructured(xml_file)
     error("Only works for structured grids ")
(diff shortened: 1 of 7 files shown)


sloede:
It seems like this PR contains the changes from #17 as well. I will thus wait with a review until #17 is merged into `main` and `main` then merged into this PR here. Feel free to ping me once this is ready for review 🙂

test/pvtk_files.jl
```diff
@@ -202,3 +230,46 @@ end
...
+  P_read = get_data_reshaped(cell_data["Pressure"], cell_data=true)
+  Phase_read = get_data_reshaped(cell_data["Phase"], cell_data=true)
+  @test P_global == P_read
+  @test Phase_global == Phase_read
+end
+
+
+# e) PVD file
+@testset "PVD" begin
+  pvd = PVDFile("examples/full_simulation.pvd")
```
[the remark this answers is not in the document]
sloede: Ah, should this maybe be something like
```suggestion
  pvd = PVDFile(joinpath(TEST_EXAMPLES_DIR, "full_simulation.pvd"))
```
?

src/ReadVTK.jl
```diff
@@ -239,7 +240,7 @@ function PVTKFile(filename; dir="")
   version = VersionNumber(attribute(root, "version", required=true))
 
   # Ensure matching file types
-  if !(file_type in ("PImageData", "PRectilinearGrid","PUnstructuredGrid"))
+  if !(file_type in ("PImageData", "PRectilinearGrid","PUnstructuredGrid", "PStructuredGrid"))
```
sloede, line 243:
```suggestion
  if !(file_type in ("PImageData", "PRectilinearGrid", "PUnstructuredGrid", "PStructuredGrid"))
```

src/ReadVTK.jl
```diff
@@ -909,11 +910,14 @@ function get_coordinates(vtk_file::VTKFile; x_string="x", y_string="y", z_string
...
-  y = get_data(coordinates[y_string])
-  z = get_data(coordinates[z_string])
+  if  vtk_file.file_type == "RectilinearGrid"
+    coordinates = get_coordinate_data(vtk_file)
+    x = get_data(coordinates[x_string])
+    y = get_data(coordinates[y_string])
+    z = get_data(coordinates[z_string])
 
+  end
+
```
sloede, line 920:
```suggestion
  coordinates = get_coordinate_data(vtk_file)
  x = get_data(coordinates[x_string])
  y = get_data(coordinates[y_string])
  z = get_data(coordinates[z_string])
```
It does not make sense to put this in an if condition, since it was already checked in lines 909ff that this mist be a file of type `RectilinearGrid`. Or am I misssing something?

src/ReadVTK.jl
```diff
@@ -928,16 +932,38 @@ Note that in VTK, points are always stored three-dimensional, even for 1D or 2D
...
+ 
+    # collect parts 
+    for i=1:length(points)
+      ex = extents[i]
+      data[1:3, ex...] .= reshape(points[i],3,length.(extents[i])...);
+      x = data[1,:,:,:]
+      y = data[2,:,:,:]
+      z = data[3,:,:,:]
+    end
+
```
sloede, line 954:
```suggestion
```
No blank lines at the end of a block

src/ReadVTK.jl
```diff
@@ -928,16 +932,38 @@ Note that in VTK, points are always stored three-dimensional, even for 1D or 2D
 See also: [`get_cells`](@ref)
 """
 function get_coordinates(pvtk_file::PVTKFile; x_string="x", y_string="y", z_string="z")
-  coords =  get_coordinates.(pvtk_file.vtk_files,x_string=x_string, y_string=y_string, z_string=z_string);
+  if pvtk_file.file_type== "PStructuredGrid"
```
sloede, line 935:
```suggestion
  if pvtk_file.file_type == "PStructuredGrid"
```

src/ReadVTK.jl
```diff
@@ -928,16 +932,38 @@ Note that in VTK, points are always stored three-dimensional, even for 1D or 2D
...
+    # collect parts 
+    for i=1:length(points)
+      ex = extents[i]
+      data[1:3, ex...] .= reshape(points[i],3,length.(extents[i])...);
+      x = data[1,:,:,:]
+      y = data[2,:,:,:]
+      z = data[3,:,:,:]
+    end
+    
+  elseif pvtk_file.file_type== "PRectilinearGrid"
```
sloede, line 955:
```suggestion
  elseif pvtk_file.file_type == "PRectilinearGrid"
```

src/ReadVTK.jl
```diff
@@ -928,16 +932,38 @@ Note that in VTK, points are always stored three-dimensional, even for 1D or 2D
...
+    coords =  get_coordinates.(pvtk_file.vtk_files,x_string=x_string, y_string=y_string, z_string=z_string);
+    x,y,z = coords[1][1][:],coords[1][2][:],coords[1][3][:]
+    for i=2:length(pvtk_file)
+      x=[x; coords[i][1][:]]
+      y=[y; coords[i][2][:]]
+      z=[z; coords[i][3][:]]
+    end
+    x,y,z = unique(x), unique(y), unique(z)
   end
```
sloede, line 964: I think there should also be an `else` condition with an error that is triggered when `get_coordinates` is called with a PVTKFile that is neither of type "PStructuredGrid" nor of type "PRectilinearGrid". Otherwise, in the current implementation you will get an error that the variables x, y, z have not been defined, which is not very helpful for the user unless they look into the implementation details

sloede (changes requested):
Thanks for the nice contribution again, @boriskaus! I have left a few comments and suggestions but nothing wild, thus I think this can be cleaned up easily for merging...

src/ReadVTK.jl
boriskaus: correct

src/ReadVTK.jl
boriskaus: fair enough; added that

test/pvtk_files.jl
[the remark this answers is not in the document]
boriskaus: yes I think that is it. Not sure why the test was not failing...

test/structuredgrid.jl
```diff
@@ -0,0 +1,146 @@
...
+  for dim in 2:3
+    # Define grid.
+    if dim == 2
+      Ni, Nj, Nk = 20, 30, 1
+
+      x = zeros(FloatType, Ni,Nj)
+      y = zeros(FloatType, Ni,Nj)
+  
+      [x[i,j] = i*i/Ni/Ni for i = 1:Ni, j = 1:Nj]
+      [y[i,j] = sqrt(j/Nj) for i = 1:Ni, j = 1:Nj]
```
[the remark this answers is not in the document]
boriskaus: yup

boriskaus:
I've implemented it all, and it seems to run fine. 
I see that you only test on Julia 1.6; perhaps an idea to also add a more recent version of Julia there (I know it should be backwards compatible, but you never know).

src/ReadVTK.jl
```diff
@@ -905,15 +906,22 @@ Note that in VTK, points are always stored three-dimensional, even for 1D or 2D
...
+    z = get_data(coordinates[z_string])
+  elseif vtk_file.file_type == "StructuredGrid"
+    wholeextent, _ = get_wholeextent(vtk_file.xml_file)
+    points = get_points(vtk_file)
+    data = reshape(points,3, wholeextent...);
+    x = data[1,:,:,:]
+    y = data[2,:,:,:]
+    z = data[3,:,:,:]
+  else
+    error("The file type of the VTK file must be 'RectilinearGrid' (current: $(vtk_file.file_type)).")
```
sloede, line 922:
```suggestion
    error("The file type of the VTK file must be 'RectilinearGrid' or 'StructuredGrid' (current: $(vtk_file.file_type)).")
```

sloede (approved):
Great work, again, LGTM! Once tests pass and coverage is ok, this can be merged...

sloede:
Ah, there seems to be an error with at least one of the tests...

boriskaus:
Hmm, I had the same error a while ago. It started when I introduced `local x,y,z` in `structuredgrid.jl` but seems to trigger an error in a completely different test. I have no idea why that would be.

boriskaus:
So [this](https://github.com/trixi-framework/ReadVTK.jl/actions/runs/4363300105/jobs/7629232436#step:6:130) mistake sometimes occurs in CI and sometimes not, even when we cleanup the files altogether. I don't like this at all...

sloede:
> So [this](https://github.com/trixi-framework/ReadVTK.jl/actions/runs/4363300105/jobs/7629232436#step:6:130) mistake sometimes occurs in CI and sometimes not, even when we cleanup the files altogether. I don't like this at all...

Can you try fixing it? It seems to me like a very specific error, not related to floating point truncation.

boriskaus:
I have no clue where to start. Seems that the files written to disc sometimes have a different size. Not having a local machine that reproduces this error does not help..

boriskaus:
Note that the main branch is also failing with what feels to be a fairly similar problem.

boriskaus:
@sloede: Whereas CI is fine now, the error remains weird and occurs when reading compressed data back from a file [here](https://github.com/trixi-framework/ReadVTK.jl/blob/c13c2c124f006843f9fba05db15dc30a837eb322/src/ReadVTK.jl#L665):
```julia
raw = data_array.vtk_file.appended_data
HeaderType = header_type(data_array.vtk_file)
if is_compressed(data_array)
   # If data is stored compressed, the first four integers of type `header_type` are the header and
   # the fourth value contains the number of bytes to read
   first = data_array.offset + 1
   last = data_array.offset + 4 * sizeof(HeaderType)
   header = Int.(reinterpret(HeaderType, raw[first:last]))
   n_bytes = header[4]

   first = data_array.offset + 4 * sizeof(HeaderType) + 1
   last = first + n_bytes - 1
   uncompressed = transcode(ZlibDecompressor, raw[first:last])
else
   ...
end
``` 
in some cases, the length of the `raw` vector is shorter than `last` as indicated in the header. The CI failure in the `main` branch seems to be caused by the same [issue](https://github.com/trixi-framework/ReadVTK.jl/actions/runs/4363069490/jobs/7628712036). 
In my last commit, I simply added an error message when this occurs, but it doesn't cure the cause of it.

Are there perhaps changes in the upstream packages that could cause this?

sloede:
> in some cases, the length of the `raw` vector is shorter than `last` as indicated in the header. The CI failure in the `main` branch seems to be caused by the same [issue](https://github.com/trixi-framework/ReadVTK.jl/actions/runs/4363069490/jobs/7628712036). In my last commit, I simply added an error message when this occurs, but it doesn't cure the cause of it.

Thanks for investigating the issue. Do you know why the error now disappeared? And did it only occur in cases where the file in question was generated on-the-fly or also in cases where an existing vtk file is used?

> Are there perhaps changes in the upstream packages that could cause this?

I believe we had a similar issue before - @andrewwinters5000 do you still remember where and when this occurred? I do not remember what we did back then to resolve it, but I recall that we had a similar issue of files being "truncated" by the zip/unzip mechanism.

test/runtests.jl
```diff
@@ -315,9 +331,14 @@ mkpath(TEST_EXAMPLES_DIR)
       @test_throws Exception get_primitives(vtk, "Foo")
       @test_throws Exception get_primitives(vtk, "Verts")
     
+      # Clean up afterwards: delete example file directory
+      clean_directory(TEST_EXAMPLES_DIR)
```
sloede, line 335:
```suggestion
```
I think it should be sufficient to recreate the test dir once per top-level testset, not in each individual testset as well

ranocha:
> > in some cases, the length of the `raw` vector is shorter than `last` as indicated in the header. The CI failure in the `main` branch seems to be caused by the same [issue](https://github.com/trixi-framework/ReadVTK.jl/actions/runs/4363069490/jobs/7628712036). In my last commit, I simply added an error message when this occurs, but it doesn't cure the cause of it.
> 
> Thanks for investigating the issue. Do you know why the error now disappeared? And did it only occur in cases where the file in question was generated on-the-fly or also in cases where an existing vtk file is used?
> 
> > Are there perhaps changes in the upstream packages that could cause this?
> 
> I believe we had a similar issue before - @andrewwinters5000 do you still remember where and when this occurred? I do not remember what we did back then to resolve it, but I recall that we had a similar issue of files being "truncated" by the zip/unzip mechanism.


I also vaguely remember some spurious issues like these. If I remember correctly, they had some stochastic part, i.e., they were not triggered every time - maybe depending on the GitHub runners etc.? Some parts may have been specific to Windows or so...

boriskaus:
> > in some cases, the length of the `raw` vector is shorter than `last` as indicated in the header. The CI failure in the `main` branch seems to be caused by the same [issue](https://github.com/trixi-framework/ReadVTK.jl/actions/runs/4363069490/jobs/7628712036). In my last commit, I simply added an error message when this occurs, but it doesn't cure the cause of it.
> 
> Thanks for investigating the issue. Do you know why the error now disappeared? 
No, if you look at the various CI's of yesterday, it sometimes appears and sometimes not (on different systems)

> And did it only occur in cases where the file in question was generated on-the-fly or also in cases where an existing vtk file is used?
I believe it is only with cases where the file was generated on-the-fly.

> 
> > Are there perhaps changes in the upstream packages that could cause this?
> 
> I believe we had a similar issue before - @andrewwinters5000 do you still remember where and when this occurred? I do not remember what we did back then to resolve it, but I recall that we had a similar issue of files being "truncated" by the zip/unzip mechanism.

sloede:
OK. Since the error seems to have disappeared, once the last remaining issues are resolved, we can merge this.

src/ReadVTK.jl
```diff
@@ -662,6 +663,11 @@ function get_data(data_array::VTKDataArray{T,N,<:FormatBinary}) where {T,N}
     last = length(raw)
 
     # Pass data through ZLib decompressor
+    if last>length(raw)
+        @show first, last, size(raw)
+        error("mistake in get_data")
+    end
```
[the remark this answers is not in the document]
boriskaus: yes indeed, this error message was in the wrong routine. Will upload a fix

boriskaus:
so we now have 1 failed CI and 2 that passed, with the only thing that changed being:
```Julia
@show data_array data_array.vtk_file.xml_file
```
and removing:
```Julia
# Start with a clean environment: remove example file directory if it exists
isdir(TEST_EXAMPLES_DIR) && rm(TEST_EXAMPLES_DIR, recursive=true)
mkpath(TEST_EXAMPLES_DIR)
```
It seems to be rather stochastic and I have no idea where to look.
I guess we can probably merge (and increment the version number) and see if it keeps occurring.

ranocha:
> It seems to be rather stochastic and I have no idea where to look.

That's also what I remember from similar issues earlier - it's stochastic and when it fails, it's on Windows.

> I guess we can probably merge (and increment the version number) and see if it keeps occurring.

Sounds good to me. Thanks a lot for your contributions!

sloede (approved):

PR #19, opened 2023-03-08
Use latest Julia release for testing

sloede:
As suggested by @boriskaus, do not just test on Julia v1.6 but also on the latest released version (semver '1'). Let's keep v1.6 around for a while, as it helps to maintain a certain level of backward comparability.


ranocha (approved):

ranocha:
Thanks! LGTM and can be merged from my point of view.

PR #22, opened 2023-03-31
Fix for connectivity to match WriteVTK output

Matthew-Whisenant:
Simple fix for connectivity to align data to match that of WriteVTK. Also fixes manual test of connectivity by incrementing by one.


Matthew-Whisenant:
Auto formatting messed this up. Fixing with another PR.

Issue #21, opened 2023-03-30
Connectivity mismatch between WriteVTK and ReadVTK


Matthew-Whisenant:
I am currently writing a function that can take VTU data read from ReadVTK and use that data to write a new VTU in WriteVTK.

As a test, I am doing the simple write VTU example on WriteVTK [here](https://jipolanco.github.io/WriteVTK.jl/dev/grids/unstructured/).

Using ReadVTK's get_cells, one can obtain the connectivity list and corresponding offset vector. However, I noticed that the connectivity list is starts at index 0 while WriteVTK starts at 1, resulting in mismatch of connectivity data. Here is a minimal example:

```julia
using ReadVTK
using WriteVTK

function convert_cell_structs(cells::VTKCells)

    start_offsets = [0; cells.offsets[1:end-1]] .+ 1
    end_offsets = cells.offsets

    offset_ranges = range.(start_offsets, end_offsets)

    connectivity = getindex.([cells.connectivity], offset_ranges)
    cell_types = VTKCellType.(cells.types)

    return MeshCell.(cell_types, connectivity)

end

points = rand(3, 5)  # 5 points in three dimensions
cells = [MeshCell(VTKCellTypes.VTK_TRIANGLE, [1, 4, 2]),
    MeshCell(VTKCellTypes.VTK_QUAD, [2, 4, 3, 5])]

vtk_grid("filename", points, cells) do vtk
    # add datasets...
end


vtkcase = VTKFile("filename.vtu")
points2 = get_points(vtkcase)
cells2 = convert_cell_structs(get_cells(vtkcase))
```

The connectivity input can be seen as
```julia
2-element Vector{MeshCell{VTKCellType, Vector{Int64}}}:
 MeshCell{VTKCellType, Vector{Int64}}(VTKCellType("VTK_TRIANGLE", 0x05, 3), [1, 4, 2])
 MeshCell{VTKCellType, Vector{Int64}}(VTKCellType("VTK_QUAD", 0x09, 4), [2, 4, 3, 5])
```
and for ReadVTK:
```julia
tempcells = get_cells(vtkcase)
VTKCells{Base.ReinterpretArray{Int64, 1, UInt8, Vector{UInt8}, false}, Base.ReinterpretArray{Int64, 1, UInt8, Vector{UInt8}, false}, Vector{UInt8}}([0, 3, 1, 1, 3, 2, 4], [3, 7], UInt8[0x05, 0x09])
```
using my function:
```julia
[...]

PR #23, opened 2023-03-31
Fix for connectivity offset from WriteVTK

Matthew-Whisenant:
Increment connectivity in get_cells by one in order to match WriteVTK output. Had to increment the manual connectivity test case by one as well.

This is a partial fix for #21.

### DIFF
--- a/src/ReadVTK.jl
+++ b/src/ReadVTK.jl
@@ -1032,7 +1032,9 @@ function get_cells(vtk_file::VTKFile)
   @assert !isnothing(offsets)
   @assert !isnothing(types)
 
-  # Create VTKCells container and convert VTK's zero-based indices to Julia's one-based indices 
+  # Create VTKCells container and convert VTK's zero-based indices to Julia's one-based indices
+  # Note: the offsets do not need to be updated since they point *past* the last entry
+  # in VTK files (C-style), while in Julia it is custom to point *at* the last entry.
   VTKCells(
     get_data(connectivity) + oneunit.(get_data(connectivity)),
     get_data(offsets),
--- a/src/ReadVTK.jl
+++ b/src/ReadVTK.jl
@@ -9,7 +9,7 @@ using LightXML: LightXML, XMLDocument, XMLElement, parse_string, attribute, has_
                 child_elements, free, content, find_element
 
 export VTKFile, VTKData, VTKDataArray, VTKCells, VTKPrimitives,           # structs
-       PVTKFile, PVTKData, PVTKDataArray, PVDFile,  
+       PVTKFile, PVTKData, PVTKDataArray, PVDFile,
        get_point_data, get_cell_data, get_data, get_data_reshaped,        # get data functions
        get_points, get_cells, get_origin, get_spacing, get_primitives,    # get geometry functions
        get_coordinates, get_coordinate_data,                              # get geometry functions
@@ -145,7 +145,7 @@ function VTKFile(filename)
   elseif file_type == "PolyData"
     piece = root[file_type][1]["Piece"][1]
     n_points = parse(Int, attribute(piece, "NumberOfPoints", required=true))
-    # TODO: decide how to handle the number of cells correctly, see  
+    # TODO: decide how to handle the number of cells correctly, see
     #       https://github.com/trixi-framework/ReadVTK.jl/pull/11
     n_cells = typemin(Int)
   end
@@ -174,7 +174,7 @@ piece(vtk_file::VTKFile) = LightXML.root(vtk_file.xml_file)[vtk_file.file_type][
 
 """
     isstructured(xml_file)
-    
+
 Returns `true` if it is a structured grid.
 """
 function isstructured(xml_file)
@@ -206,11 +206,11 @@ Hold all relevant information about a Parallel VTK XML file that has been read i
 - `vtk_files`: vector with `VTKFile` data that contains the info about each of the files
 """
 struct PVTKFile
-  filename::String     
+  filename::String
   xml_file::XMLDocument
   file_type::String
   version::VersionNumber
-  vtk_filenames::Vector{String} 
+  vtk_filenames::Vector{String}
   vtk_files::Vector{VTKFile}
 end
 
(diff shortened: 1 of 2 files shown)


src/ReadVTK.jl
```diff
@@ -1032,8 +1032,12 @@ function get_cells(vtk_file::VTKFile)
   @assert !isnothing(offsets)
   @assert !isnothing(types)
 
-  # Create VTKCells container
-  VTKCells(get_data(connectivity), get_data(offsets), get_data(types))
+  # Create VTKCells container, note that connectivity must be offset by one to match WriteVTK data
+  VTKCells(
+    get_data(connectivity) + oneunit.(get_data(connectivity)),
+    get_data(offsets),
```
sloede, line 1038: I think the offsets need to be incremented by one as well, don't they?
```suggestion
    get_data(offsets) + oneunit.(get_data(connectivity)),
```

src/ReadVTK.jl
```diff
@@ -1032,8 +1032,12 @@ function get_cells(vtk_file::VTKFile)
   @assert !isnothing(offsets)
   @assert !isnothing(types)
 
-  # Create VTKCells container
-  VTKCells(get_data(connectivity), get_data(offsets), get_data(types))
+  # Create VTKCells container, note that connectivity must be offset by one to match WriteVTK data
```
sloede, line 1035:
```suggestion
  # Create VTKCells container and convert VTK's zero-based indices to Julia's one-based indices 
```

test/runtests.jl
```diff
@@ -100,7 +100,7 @@ clean_directory(TEST_EXAMPLES_DIR) = @test_nowarn rm(TEST_EXAMPLES_DIR, recursiv
       @test size(cells) == (3085,)
       @test div(Int(sum(cells.types)), 8) == 3085
       @test sum(cells.offsets) == 19040620
```
sloede, line 102: If applying the offset fix above, this value probably needs changing as well.

src/ReadVTK.jl
sloede: But please check if that makes sense. I believe that the offsets should point at the last value for each cell, thus the final offset should probably be the same as the length of the points array?

Matthew-Whisenant:
Ah well. Clearly something with my whitespace formatting is still auto formatting. I'll leave this one up and try to figure out how I can eliminate that in the future.

sloede (changes requested):
Great, thanks a lot for the the contribution. I've left a few minor comments, but once resolved, this can be merged.

src/ReadVTK.jl
Matthew-Whisenant: The change you suggest would be like:
```julia
get_data(offsets) + oneunit.(get_data(offsets))
```


and working through a simple example like:


```julia
points = rand(3, 5)  # 5 points in three dimensions
cells = [MeshCell(VTKCellTypes.VTK_TRIANGLE, [1, 4, 2]),
    MeshCell(VTKCellTypes.VTK_QUAD, [2, 4, 3, 5])]

vtk_grid("test/simplevtk", points, cells) do vtk
    # add datasets...
end

vtkcase = VTKFile("test/simplevtk.vtu")
points = get_points(vtkcase)
cells = get_cells(vtkcase)
```

from cells you get 
```julia
VTKCells{Vector{Int64}, Vector{Int64}, Vector{UInt8}}([1, 4, 2, 2, 4, 3, 5], [4, 8], UInt8[0x05, 0x09])
```
This offset points to 4 and 8, which the length of the connectivity vector is 7 anyway. So this seems wrong. Without offsetting the offset array I get:
```julia
VTKCells{Vector{Int64}, Base.ReinterpretArray{Int64, 1, UInt8, Vector{UInt8}, false}, Vector{UInt8}}([1, 4, 2, 2, 4, 3, 5], [3, 7], UInt8[0x05, 0x09])
```
Which seems correct, although now that I see this, I notice that the oneunit operation changes the connectivity from type ReinterpretArray to a proper Vector, though I don't think this changes again significant.

Without any changes in this PR the current output would be:
```julia
VTKCells{Base.ReinterpretArray{Int64, 1, UInt8, Vector{UInt8}, false}, Base.ReinterpretArray{Int64, 1, UInt8, Vector{UInt8}, false}, Vector{UInt8}}([0, 3, 1, 1, 3, 2, 4], [3, 7], UInt8[0x05, 0x09])
```

src/ReadVTK.jl
Matthew-Whisenant: Sure. Not sure if the "Resolve converstion" accepts the change but yes.

sloede: No, it doesn't :-) You have to press "Commit suggestion", which will auto-resolve the convo. I'll leave it up to you to do this here.

src/ReadVTK.jl
sloede: I think you're right. It seems like the offset is "into the connectivity array for the end of each cell" (https://kitware.github.io/vtk-examples/site/VTKFileFormats/#unstructuredgrid), which imho is ambiguous, but which seems to be C-style. That is, it points to *one **past** the last point* in the connectivity array. I'll close this and the remark below as resolved.

test/runtests.jl
sloede: Not necessary, see convo above.

src/ReadVTK.jl
```diff
@@ -1032,8 +1032,12 @@ function get_cells(vtk_file::VTKFile)
   @assert !isnothing(offsets)
   @assert !isnothing(types)
 
-  # Create VTKCells container
-  VTKCells(get_data(connectivity), get_data(offsets), get_data(types))
+  # Create VTKCells container and convert VTK's zero-based indices to Julia's one-based indices
```
sloede, line 1035:
```suggestion
  # Create VTKCells container and convert VTK's zero-based indices to Julia's one-based indices
  # Note: the offsets do not need to be updated since they point *past* the last entry
  # in VTK files (C-style), while in Julia it is custom to point *at* the last entry.
```

sloede (approved):
LGTM and ready to merge once all tests pass

sloede:
The Windows tests are failing due to a know (but intermittent) issue with decompression, thus I'll go forward and merge this manually.

PR #26, opened 2023-04-05
update to VTKBase.jl

ranocha:
This is just the basic template for the update to use types from VTKBase.jl. Building the docs works and we can use this as a starting point for further extensions.


sloede (approved):
LGTM, thanks!

PR #27, opened 2023-04-05
do not create empty function just for setting a docstring

ranocha:
The previous version created a method of `get_data` that just returns `nothing`.


sloede (approved):
Thanks for catching this!

Issue #20, opened 2023-03-28
Regarding OpenFOAM VTK Files


Matthew-Whisenant:
I just wanted to comment on the types of VTK files that ReadVTK.jl can read.

My current workflow is:
run OpenFOAM -> Paraview to convert foam to vtu -> ReadVTK to load vtu

This works great for everything that is not a scalar. However, OpenFOAM's vtk writer apparently does not include the "NumberOfComponents" tag if a field is a scalar.

On line 610 of ReadVTK we have:
```julia 
n_components = parse(Int, attribute(xml_element, "NumberOfComponents", required=true))
```

This requires that field. However, by replacing this line with:
```julia
n_components =
    try
      parse(Int, attribute(xml_element, "NumberOfComponents", required=true))
    catch
      1
    end
```
Then everything works fine. Scalars loaded and I don't see anything loaded incorrectly.

Perhaps this could be added (probably a better way then what I have done here) and then advertised that this can now read OpenFOAM data. I know the CFD field is increasing using Julia so this might be a boon for this repository.

ranocha:
I think it is a good idea in general to extend this package. However, I would not like to use exception handling for this case - something like `has_attribute(xml_element, "NumberOfComponents")` should be better, I think.

You are welcome to prepare a PR with the change (including a test)!

Matthew-Whisenant:
That sounds good. I am a PhD candidate in school so I do not have much experience doing PRs and open source collaboration in general. But its a good thing to figure out. 

Since I can't reproduce a test with WriteVTK, I assume I can do a PR for the examples repo with my included Paraview vtu?

Is it also correct to assume you mean a test for the 'test' command in the package manager?

ranocha:
> > Since I can't reproduce a test with WriteVTK, I assume I can do a PR for the examples repo with my included Paraview vtu?
>
> Yeah, that's a bit tricky. You could also generate a test file with WriteVTK.jl (or copy an existing one) and then remove the NumberOfComponents attribute.

Yes, that's basically what I had in mind - write your own file for a very simple case. Using WriteVTK.jl and our dependency LightXML.jl should enable you to create a file that mimics the output you get from your standard workflow.

Matthew-Whisenant:
So I've been working on this, and unfortunately the ParaView vtu format is just too different from WriteVTK to be a good test case. I had to change a line in get_data in order to get the non-appended binary data read in correctly. 

Essentially this line
[...]

PR #28, opened 2023-04-07
Added ParaView file output read ability.

Matthew-Whisenant:
This PR adds the lines needed to read the slightly different ParaView output format of VTKs. The base64decode needed another filter for the raw data. ParaView also doesn't give the "NumberOfComponents" tag to scalars, so that line was changed per issue #20 

Also added some tests that read to files with identical data, but one was read and then re-output by ParaView. I've submitted a PR for that test file in the examples repo. 

Lastly the "convert_VTKCells_to_MeshCells" does what it says. It can be named something better, but at least its here to show how to convert between VTKCells style to MeshCells. 

I realized after the first commit here, that this function probably should have been a separate PR. Sorry about that, still learning the ropes. I can redo everything if necessary.

Please let me know any concerns or suggestions you have.

### DIFF
--- a/src/ReadVTK.jl
+++ b/src/ReadVTK.jl
@@ -16,7 +16,8 @@ export VTKFile, VTKData, VTKDataArray, VTKCells, VTKPrimitives,           # stru
        get_point_data, get_cell_data, get_data, get_data_reshaped,        # get data functions
        get_points, get_cells, get_origin, get_spacing, get_primitives,    # get geometry functions
        get_coordinates, get_coordinate_data,                              # get geometry functions
-       get_example_file                                                   # other functions
+       get_example_file,                                                  # other functions
+       convert_VTKCells_to_MeshCells                                      # conversion utility
 
 """
     VTKFile
@@ -1137,6 +1138,29 @@ function get_example_file(filename; head="main", output_directory=".", force=fal
   end
 
   return filepath
+end
+
+"""
+    convert_VTKCells_to_MeshCells(cells::VTKCells)
+
+Converts the `VTKCells` type, which holds data on N number of cells, to
+a vector of N number of `MeshCell` types. Allows for the reading and then
+subsequent writing with `WriteVTK`.
+
+See also: [`VTKCells`](@ref), [`MeshCell`](@ref)
+"""
+function convert_VTKCells_to_MeshCells(cells::VTKCells)
+
+  start_offsets = [0; cells.offsets[1:end-1]] .+ 1
+  end_offsets = cells.offsets
+
+  offset_ranges = range.(start_offsets, end_offsets)
+
+  connectivity = getindex.([cells.connectivity], offset_ranges)
+  cell_types = VTKBase.VTKCellType.(cells.types)
+
+  return VTKBase.MeshCell.(cell_types, connectivity)
+
 end
 
 end # module
--- a/src/ReadVTK.jl
+++ b/src/ReadVTK.jl
@@ -610,8 +610,14 @@ function VTKDataArray(xml_element, vtk_file::VTKFile)
   # Extract information about the underlying data
   data_type = string_to_data_type(attribute(xml_element, "type", required=true))
   name = attribute(xml_element, "Name", required=true)
-  n_components = parse(Int, attribute(xml_element, "NumberOfComponents", required=true))
   format_string = attribute(xml_element, "format", required=true)
+
+  # ParaView doesn't include NumberOfComponents tag if field is a scalar
+  if has_attribute(xml_element, "NumberOfComponents")
+    n_components = parse(Int, attribute(xml_element, "NumberOfComponents", required=true))
+  else
+    n_components = 1
+  end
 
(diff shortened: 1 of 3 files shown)


Matthew-Whisenant:
After this both #20 and #21 can probably be closed.

src/ReadVTK.jl
```diff
@@ -657,7 +664,8 @@ function get_data(data_array::VTKDataArray{T,N,<:FormatBinary}) where {T,N}
   # * first get the content of the corresponding XML data array
   # * then remove leading/trailing whitespace
   # * finally decode from Base64 to binary representation
-  raw = base64decode(strip(content(data_array.data_array)))
+  # * split with "\n" added to read ParaView output files
+  raw = base64decode(split(strip(content(data_array.data_array)), "\n")[1])
```
sloede, line 668: What exactly is happening here? It seems to me like the content is split upon `\n` and then only the first entry is returned. What kind of data does ParaView put after the `\n` that should be ignored?

src/ReadVTK.jl
```diff
@@ -1132,4 +1140,27 @@ function get_example_file(filename; head="main", output_directory=".", force=fal
   return filepath
 end
 
+"""
+    convert_VTKCells_to_MeshCells(cells::VTKCells)
+
+Converts the `VTKCells` type, which holds data on N number of cells, to
+a vector of N number of `MeshCell` types. Allows for the reading and then
+subsequent writing with `WriteVTK`.
```
sloede, line 1148:
```suggestion
Convert a `VTKCells` object, which holds raw point and connectivity data for a number of cells, to
a vector of `MeshCell` objects. The latter can, e.g., be passed to the WriteVTK.jl package for writing new VTK files.
```
We are not really converting types but objects of a given type; I hope this makes the comment a little clearer.

src/ReadVTK.jl
```diff
@@ -1132,4 +1140,27 @@ function get_example_file(filename; head="main", output_directory=".", force=fal
   return filepath
 end
 
+"""
+    convert_VTKCells_to_MeshCells(cells::VTKCells)
```
sloede, line 1144: I am not yet convinced that this is the best name we can come up with. Maybe just `convert_to_meshcells`? Do you have an idea, @ranocha or @jipolanco?

sloede (commented):
In general this looks already quite good to me. Please see the attached comments/suggestion for more details.

Some thoughts:
* It seems reasonable for me that the user should call a conversion function explicitly if they want to access the cell information in a WriteVTK.jl-compatible manner. Or does anyone think this should be the default?
* It would be great if you could add an example for this explicit usage to the docs, such that users are aware of this function right away.
* You should update the README and state something like "Some ParaView-created files can be read (still experimental)"
* You should also add your name to the authors section.

src/ReadVTK.jl
```diff
@@ -1132,4 +1140,27 @@ function get_example_file(filename; head="main", output_directory=".", force=fal
...
+
+  start_offsets = [0; cells.offsets[1:end-1]] .+ 1
+  end_offsets = cells.offsets
+
+  offset_ranges = range.(start_offsets, end_offsets)
+
+  connectivity = getindex.([cells.connectivity], offset_ranges)
+  cell_types = VTKBase.VTKCellType.(cells.types)
+
+  return VTKBase.MeshCell.(cell_types, connectivity)
```
ranocha, line 1162: I guess this will require #26 since we do not depend on VTKBase.jl right now?

src/ReadVTK.jl
jipolanco: Yeah, I'd suggest something like `to_meshcells` or `to_writevtk`.

src/ReadVTK.jl
sloede: Right. Could you turn #26 into a proper PR such that we can start merging it?

src/ReadVTK.jl
sloede: Let's use `to_meshcells` for now (since it is the most descriptive). If we find other instances of conversions for which it would make sense to have a single `to_writevtk` API, we can always `@deprecate` it...

src/ReadVTK.jl
ranocha: Done

sloede:
Why was this closed when I merged #26? @Matthew-Whisenant please feel free to reopen this, this closing it was not my intent.

Matthew-Whisenant:
I don't exactly see how to reopen. Maybe cause the branch already merged? I can reopen on the main branch. Should we move the conversation to that PR or continue on this one?

sloede:
Ah, right. If you press the `Edit` button above next to the PR title, you should be able to change the merge target to `main`. Then, you should be able to reopen this PR

sloede:
> I have also clicked the "Allow edits by maintainers" if that, for some reason, allows you to reopen or do something. If it does work I'll just update this PR.

Sorry, thus didn't to the trick. You will have to recreate a new PR.

PR #30, opened 2023-04-16
Adds VTKCells converter

Matthew-Whisenant:
This PR follows up on the accidentally closed PR #28 

This PR sets up the workflow between ReadVTK -> WriteVTK by adding `to_meshcell`. Here the changes are:

- New function `to_meshcell` that converts `VTKCells` type to multiple WriteVTK compliant `MeshCells`
- Test in the first basic testing to check for `to_meshcell` correctness
- Added example of workflow in main index page. Let me know if you want something changed.
- Added my name to contributors. My university page got cleared in a address move but will be back up in a week or so. This new address will be the permanent one so no need for a recommit.
- Added .vscode to gitignore to help vscode users settings not mess with commits. Helps me not added whitespace to commits.

Note that I separated PRs in this to make more sense on an orthogonal feature level. My next PR will include more inclusive testing of ReadVTK -> WriteVTK workflow with ParaView and standard files. But this single test should at least cover testing of the conversion in and of itself.

This PR must be accepted before the larger ParaView one, but I figured it was worth it for clarity to split them up.

PR #31, opened 2023-04-16
ParaView read ability

Matthew-Whisenant:
See #28 

Edit: for some reason I expected the other PR commits to auto-resolve and not be included in this PR. Maybe after they other PR is accepted?


sloede:
> for some reason I expected the other PR commits to auto-resolve and not be included in this PR. Maybe after they other PR is accepted?

I think there might be some conflicts you'll need to resolve. However, since this PR seems to depend on #30, I will wait with a review until #31 is merged (which imho can be very soon)

Matthew-Whisenant:
Everything should be in order now. Now that I know how the formatter works and is intergraded into vscode it seems like a great addition.

Again, PR depends on https://github.com/JuliaVTK/ReadVTK_examples/pull/1

All concerns should be address from previous PR https://github.com/JuliaVTK/ReadVTK.jl/pull/28

> * It seems reasonable for me that the user should call a conversion function explicitly if they want to access the cell information in a WriteVTK.jl-compatible manner. Or does anyone think this should be the default?

It would make sense if it was default to me, but that's probably for a bigger rebase if/when the exported types from both ReadVTK and WriteVTK are hammered down, such as `VTKFile` such that both add methods to existing types from VTKBase. At that point, it would also just make sense to have a central `VTKIO` package that imports both write and read, much like DifferentialEquations.jl and its library of DiffEq packages.

But that's a lot of work and I think the current way is perfectly fine.

Matthew-Whisenant:
https://github.com/JuliaVTK/ReadVTK.jl/blob/ec581dcfac5ffd54b3fcd1887cc21c8e2a662c72/test/runtests.jl#L5-L10

I suppose this needs to be updated to:
1178cafb12bfac2a6d249f3ef0da9423f6c7202e
so that it pulls the latest repo?

sloede:
> https://github.com/JuliaVTK/ReadVTK.jl/blob/ec581dcfac5ffd54b3fcd1887cc21c8e2a662c72/test/runtests.jl#L5-L10
> 
> I suppose this needs to be updated to: 1178cafb12bfac2a6d249f3ef0da9423f6c7202e so that it pulls the latest repo?

Yes, I believe so 👍

src/ReadVTK.jl
```diff
@@ -670,7 +676,8 @@ function get_data(data_array::VTKDataArray{T, N, <:FormatBinary}) where {T, N}
   # * first get the content of the corresponding XML data array
   # * then remove leading/trailing whitespace
   # * finally decode from Base64 to binary representation
-  raw = base64decode(strip(content(data_array.data_array)))
+  # * split with "\n" added to read ParaView output files
```
sloede, line 679: This comment should go before the one above, i.e., between "then remove..." and "finally decode...".

Also, could you maybe explain what's happening here, i.e., why is there an extra `\n` and why is only the first part of the split required to be read?

Matthew-Whisenant: This is because of how ParaView output format is. In my example ParaView file, most fields don't have extra stuff added to them, but noticeably, points does. Sometimes ParaView will put tags such as 

https://github.com/JuliaVTK/ReadVTK_examples/blob/main/examples/celldata_inline_binary_uncompressed_ParaView.vtu#L26-L43

at the end of some fields, which I assume is data used for data coloring so it doesn't have to be recalculated every time.  Here is the last couple lines in the binary read with just strip/content:

```
BIQAAAAAAAAFBAAAAAAAAAAAAAAAAAAABQQAAAAAAAAFBAAAAAAAAAAAA=
          
            
              0
            
            
              90.509667992
            
          
          
            
              0
            
            
              90.509667992
```

This data is easily delimited by  "\n" , else the get_data function will get the numbers here as well as the inline binary.
For the file I used in testing by exporting from ParaView, only points field had this. But in general, any field could have this.
I believe this is also where ParaView will add extra data regarding a field in general, not just L2_NORM. 

I suspect this is why appended data reading of ParaView files does not work, but the the fix seemed much more involved.

Matthew-Whisenant: Also I swapped the comments in the file as you said.

src/ReadVTK.jl
```diff
@@ -669,8 +675,9 @@ function get_data(data_array::VTKDataArray{T, N, <:FormatBinary}) where {T, N}
   # To retrieve the raw data,
   # * first get the content of the corresponding XML data array
   # * then remove leading/trailing whitespace
+  # * split with "\n" added to read ParaView output files
```
sloede, line 678:
```suggestion
  # * split with "\n" added to read ParaView output files
  #   Rationale: In "binary" format, ParaView sometimes adds additional XML tags with
  #              further information after the data itself. It seems like these additional
  #              tags are always separated by a newline, thus we `split` the input on newlines
  #              and only consider the content before the first newline.
  #              Example: https://github.com/JuliaVTK/ReadVTK_examples/blob/1178cafb12bfac2a6d249f3ef0da9423f6c7202e/examples/celldata_inline_binary_uncompressed_ParaView.vtu#L26-L43
  #              Details: https://github.com/JuliaVTK/ReadVTK.jl/pull/31/files#r1174125579
```
I would like to preserve this information for future me 🙂

src/ReadVTK.jl
sloede: Great, thanks for the detailed information!

sloede (approved):
LGTM! Thanks again for this nice contribution!

Matthew-Whisenant:
Thanks, I've closed #21 and #20 because this was merged.

sloede:
👍 Thanks! Please feel welcome to further improve the capabilities of ReadVTK.jl any time 🙂

PR #32, opened 2023-04-20
Add JuliaFormatter and format `src`

ranocha:
This enables JuliaFormatter and applies it to `src`. This should lead to a consistent formatting and make it easier to maintain given the number of PRs from various external contributors.

As a next step, we should
- format the remaining folders
- setup a CI check

I would like to do this in another PR to keep the amount of code changes per PR low enough to make it easier to review.

### DIFF
--- a/src/ReadVTK.jl
+++ b/src/ReadVTK.jl
@@ -57,7 +57,7 @@ mutable struct VTKFile
     f(vtk_file) = free(vtk_file.xml_file)
 
     # Register finalizer
-    finalizer(f, vtk_file)
+    return finalizer(f, vtk_file)
   end
 end
 
@@ -96,7 +96,8 @@ function VTKFile(filename)
     offset_begin = first(findnext("_", raw_file_contents, last(marker))) + 1
     offset_end = first(findnext("</AppendedData>", raw_file_contents, offset_begin)) - 1
     appended_data = Vector{UInt8}(rstrip(raw_file_contents[offset_begin:offset_end]))
-    xml_file_contents = raw_file_contents[1:offset_begin-1] * "\n  </AppendedData>\n</VTKFile>"
+    xml_file_contents = raw_file_contents[1:(offset_begin - 1)] *
+                        "\n  </AppendedData>\n</VTKFile>"
   end
 
   # Open file and ensure that it is a valid VTK file
@@ -105,18 +106,19 @@ function VTKFile(filename)
   @assert LightXML.name(root) == "VTKFile"
 
   # Extract attributes (use `required=true` to fail fast & hard in case of unexpected content)
-  file_type = attribute(root, "type", required=true)
-  version = VersionNumber(attribute(root, "version", required=true))
-  byte_order = attribute(root, "byte_order", required=true)
-  header_type = attribute(root, "header_type", required=true)
+  file_type = attribute(root, "type", required = true)
+  version = VersionNumber(attribute(root, "version", required = true))
+  byte_order = attribute(root, "byte_order", required = true)
+  header_type = attribute(root, "header_type", required = true)
   if has_attribute(root, "compressor")
-    compressor = attribute(root, "compressor", required=true)
+    compressor = attribute(root, "compressor", required = true)
   else
     compressor = ""
   end
 
   # Ensure matching file types
-  if !(file_type in ("UnstructuredGrid", "ImageData", "PolyData", "RectilinearGrid", "StructuredGrid"))
+  if !(file_type in ("UnstructuredGrid", "ImageData", "PolyData", "RectilinearGrid",
+                     "StructuredGrid"))
     error("Unsupported file type: ", file_type)
   end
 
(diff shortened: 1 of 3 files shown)


src/ReadVTK.jl
```diff
@@ -96,7 +94,8 @@ function VTKFile(filename)
     offset_begin = first(findnext("_", raw_file_contents, last(marker))) + 1
     offset_end = first(findnext("</AppendedData>", raw_file_contents, offset_begin)) - 1
     appended_data = Vector{UInt8}(rstrip(raw_file_contents[offset_begin:offset_end]))
-    xml_file_contents = raw_file_contents[1:offset_begin-1] * "\n  </AppendedData>\n</VTKFile>"
+    xml_file_contents = raw_file_contents[1:(offset_begin - 1)] *
+                        "\n  </AppendedData>\n</VTKFile>"
```
sloede, line 98: I don't like this - we've had hard-to-find bugs because of implicit line continuations before. Is there a way to force the use of parentheses if a line is wrapped over multiple lines?

src/ReadVTK.jl
```diff
@@ -137,43 +137,48 @@ function VTKFile(filename)
   if file_type == "UnstructuredGrid"
     # Extract piece
     piece = root[file_type][1]["Piece"][1]
-    n_points = parse(Int, attribute(piece, "NumberOfPoints", required=true))
-    n_cells = parse(Int, attribute(piece, "NumberOfCells", required=true))
-  elseif file_type == "ImageData" || file_type == "RectilinearGrid" || file_type == "StructuredGrid"
+    n_points = parse(Int, attribute(piece, "NumberOfPoints", required = true))
+    n_cells = parse(Int, attribute(piece, "NumberOfCells", required = true))
+  elseif file_type == "ImageData" || file_type == "RectilinearGrid" ||
+         file_type == "StructuredGrid"
```
sloede, line 143: As above...

src/ReadVTK.jl
```diff
@@ -137,43 +137,48 @@ function VTKFile(filename)
...
     dataset_element = root[file_type][1]
-    whole_extent = parse.(Int, split(attribute(dataset_element, "WholeExtent", required=true), ' '))
-    n_points_per_grid_dir = [whole_extent[2*i]+1 for i in (1:3)]
+    whole_extent = parse.(Int,
+                          split(attribute(dataset_element, "WholeExtent", required = true),
+                                ' '))
```
sloede, line 147: This is just ugly,  but I guess one can live with it

src/ReadVTK.jl
```diff
@@ -224,7 +228,7 @@ Read in and parse the PVTK XML file specified by its `filename`.
 Optionally, an additional directory name `dir` can be specified for the
 location of the underlying (serial) VTK files.
 """
-function PVTKFile(filename; dir="")
+function PVTKFile(filename; dir = "")
```
sloede, line 231: Our prevalent style in Trixi.jl was to not use spaces around kwarg `=`. Should we maybe adapt this?

src/ReadVTK.jl
```diff
@@ -377,7 +381,6 @@ end
 # Reduce REPL noise by defining `show`
 Base.show(io::IO, data::PVTKData) = print(io, "PVTKData()")
 
-
```
sloede, line 380: Is there a way to keep extra blank lines? Personally, I really like the "one blank line for code separation, two blank lines for function separation" rule such that it is easier to visually track the end of a function and the beginning of a new one when scrolling.

src/ReadVTK.jl
```diff
@@ -137,43 +137,48 @@ function VTKFile(filename)
...
+        "\"", vtk_file.compressor, "\", ",
+        "<appended_data>", ", ",
+        vtk_file.n_points, ", ",
+        vtk_file.n_cells, ")")
 end
 
 # Return `Piece` XML element that contains all VTK data
-piece(vtk_file::VTKFile) = LightXML.root(vtk_file.xml_file)[vtk_file.file_type][1]["Piece"][1]
+function piece(vtk_file::VTKFile)
+  LightXML.root(vtk_file.xml_file)[vtk_file.file_type][1]["Piece"][1]
```
sloede, line 180:
```suggestion
  return LightXML.root(vtk_file.xml_file)[vtk_file.file_type][1]["Piece"][1]
```

src/ReadVTK.jl
```diff
@@ -466,7 +467,9 @@ Retrieve a lightweight `{VTKData` object with the coordinate data of the given V
 
 See also: [`PVTKData`](@ref), [`get_point_data`](@ref),  [`get_cell_data`](@ref)
 """
-get_coordinate_data(pvtk_file::PVTKFile) = get_data_section.(pvtk_file.vtk_files, "Coordinates")
+function get_coordinate_data(pvtk_file::PVTKFile)
+  get_data_section.(pvtk_file.vtk_files, "Coordinates")
```
sloede, line 471:
```suggestion
  return get_data_section.(pvtk_file.vtk_files, "Coordinates")
```
Is there a way to automatically add `return` of the function is reformatted this way?

src/ReadVTK.jl
```diff
@@ -688,7 +688,7 @@ function get_data(data_array::VTKDataArray{T,N,<:FormatBinary}) where {T,N}
 end
 
 # Retrieve actual data for XML data array (version for storage format "appended")
-function get_data(data_array::VTKDataArray{T,N,<:FormatAppended}) where {T,N}
+function get_data(data_array::VTKDataArray{T, N, <:FormatAppended}) where {T, N}
```
sloede, line 691: I guess this too is something that is a deviation from the prevalent style we use in Trixi.jl. However, here I am not as married to the status quo as for the whitespace around kwarg `=` 😬, so if nobody objects, I'd be fine with it.

src/ReadVTK.jl
```diff
@@ -788,49 +787,49 @@ function get_data_reshaped(data_array::PVTKDataArray; cell_data=false)
...
+  if N == 1
+    data = zeros(type, wholeextent...)
   else
-    data = zeros(type,N, wholeextent...)
+    data = zeros(type, N, wholeextent...)
   end
 
   # collect parts
-  for i=1:length(data_array.data)
+  for i in 1:length(data_array.data)
```
sloede, line 797: ❤️

src/ReadVTK.jl
```diff
@@ -788,49 +787,49 @@ function get_data_reshaped(data_array::PVTKDataArray; cell_data=false)
...
     ex = extents[i]
     if cell_data
-      ex =  (ex[1][1:end-1], ex[2][1:end-1], ex[3][1:end-1])
+      ex = (ex[1][1:(end - 1)], ex[2][1:(end - 1)], ex[3][1:(end - 1)])
```
sloede, line 800: ❤️ ❤️

src/get_functions.jl
```diff
@@ -86,8 +88,9 @@ Retrieve a vector with the `WholeExtent` 6-entry vector from the uniform grid [`
 
 See also: [`PVTKFile`](@ref)
 """
-get_whole_extent(pvtk_file::PVTKFile) = get_whole_extent(VTKFile(pvtk_file.vtk_filenames[1]))
-
+function get_whole_extent(pvtk_file::PVTKFile)
+  get_whole_extent(VTKFile(pvtk_file.vtk_filenames[1]))
```
sloede, line 92:
```suggestion
  return get_whole_extent(VTKFile(pvtk_file.vtk_filenames[1]))
```

src/ReadVTK.jl
ranocha: I updated the settings and re-ran the formatter from scratch.

src/ReadVTK.jl
ranocha: Yes - done

src/ReadVTK.jl
ranocha: I couldn't find something like this in the docs and filed an issue: https://github.com/domluna/JuliaFormatter.jl/issues/709

sloede: Do you know what would happen if we add the parens ourselves and then run the formatter again? Would it keep the parens or remove them?

src/ReadVTK.jl
sloede: I have no strong opinion here, so maybe let's just leave it for now as it is done in this PR (i.e., with whitespaces)

src/ReadVTK.jl
ranocha: 5b35ae1

src/ReadVTK.jl
ranocha: It keeps them: 5b35ae1

sloede: Great!

ranocha:
Windows CI failure is spurious?

PR #34, opened 2023-05-05
Bugfix - reading appended_data

boriskaus:
This fixes a bug that occurred irregularly (with files that could be read fine by Paraview). 
The culprit turns out to be the following line in `VTKFile(filename)`:
```Julia
appended_data = Vector{UInt8}(rstrip(raw_file_contents[offset_begin:offset_end]))
```
The issue is that `raw_file_contents` sometimes ends in a space as in `(f\x12 \n\t"` which `rstrip` changes to `(f\x12"` instead of `(f\x12 "`.
Fixed with:
```Julia
appended_data = Vector{UInt8}(raw_file_contents[offset_begin:offset_end][1:end-2])
```
Here a ZIP-file that has the problem:
[Convection_p00000000.vtr.zip](https://github.com/JuliaVTK/ReadVTK.jl/files/11411654/Convection_p00000000.vtr.zip)
reproduce the mistake with:
```Julia
julia> vtk_read = VTKFile("Convection_p00000000.vtr");
julia> point_data = get_point_data(vtk_read);
julia> get_data_reshaped(point_data[point_data.names[9]]);
ERROR: BoundsError: attempt to access 738759-element Vector{UInt8} at index [671681:738760]
Stacktrace:
 [1] throw_boundserror(A::Vector{UInt8}, I::Tuple{UnitRange{Int64}})
   @ Base ./abstractarray.jl:703
 [2] checkbounds
   @ ./abstractarray.jl:668 [inlined]
 [3] view
   @ ./subarray.jl:177 [inlined]
 [4] get_data(data_array::VTKDataArray{Float32, 1, ReadVTK.FormatAppended})
   @ ReadVTK ~/Downloads/ReadVTK.jl/src/ReadVTK.jl:756
 [5] #get_data_reshaped#5
   @ ~/Downloads/ReadVTK.jl/src/ReadVTK.jl:792 [inlined]
 [6] get_data_reshaped(data_array::VTKDataArray{Float32, 1, ReadVTK.FormatAppended})
   @ ReadVTK ~/Downloads/ReadVTK.jl/src/ReadVTK.jl:791
 [7] top-level scope
   @ REPL[11]:1
```

### DIFF
--- a/src/ReadVTK.jl
+++ b/src/ReadVTK.jl
@@ -96,7 +96,13 @@ function VTKFile(filename)
     # library
     offset_begin = first(findnext("_", raw_file_contents, last(marker))) + 1
     offset_end = first(findnext("</AppendedData>", raw_file_contents, offset_begin)) - 1
-    appended_data = Vector{UInt8}(rstrip(raw_file_contents[offset_begin:offset_end]))
+  
+    #appended_data = Vector{UInt8}(rstrip(raw_file_contents[offset_begin:offset_end]))
+    # rstrip was used earlier, but removes a space at the end of the line, which can cause havoc 
+    # in case raw_file_contents ends with a space, such as 
+    # "... (f\x12 \n\t", which will be turned into "... (f\x12" 
+    appended_data = Vector{UInt8}(raw_file_contents[offset_begin:offset_end][1:end-2])
+    
     xml_file_contents = (raw_file_contents[1:(offset_begin - 1)] *
                          "\n  </AppendedData>\n</VTKFile>")
   end
--- a/src/ReadVTK.jl
+++ b/src/ReadVTK.jl
@@ -96,13 +96,13 @@ function VTKFile(filename)
     # library
     offset_begin = first(findnext("_", raw_file_contents, last(marker))) + 1
     offset_end = first(findnext("</AppendedData>", raw_file_contents, offset_begin)) - 1
-  
+
     #appended_data = Vector{UInt8}(rstrip(raw_file_contents[offset_begin:offset_end]))
     # rstrip was used earlier, but removes a space at the end of the line, which can cause havoc 
     # in case raw_file_contents ends with a space, such as 
     # "... (f\x12 \n\t", which will be turned into "... (f\x12" 
-    appended_data = Vector{UInt8}(raw_file_contents[offset_begin:offset_end][1:end-2])
-    
+    appended_data = Vector{UInt8}(raw_file_contents[offset_begin:offset_end][1:(end - 2)])
+
     xml_file_contents = (raw_file_contents[1:(offset_begin - 1)] *
                          "\n  </AppendedData>\n</VTKFile>")
   end
--- a/src/ReadVTK.jl
+++ b/src/ReadVTK.jl
@@ -96,12 +96,7 @@ function VTKFile(filename)
     # library
     offset_begin = first(findnext("_", raw_file_contents, last(marker))) + 1
     offset_end = first(findnext("</AppendedData>", raw_file_contents, offset_begin)) - 1
-
-    #appended_data = Vector{UInt8}(rstrip(raw_file_contents[offset_begin:offset_end]))
-    # rstrip was used earlier, but removes a space at the end of the line, which can cause havoc 
-    # in case raw_file_contents ends with a space, such as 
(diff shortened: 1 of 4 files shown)


src/ReadVTK.jl
```diff
@@ -96,7 +96,13 @@ function VTKFile(filename)
     # library
     offset_begin = first(findnext("_", raw_file_contents, last(marker))) + 1
     offset_end = first(findnext("</AppendedData>", raw_file_contents, offset_begin)) - 1
-    appended_data = Vector{UInt8}(rstrip(raw_file_contents[offset_begin:offset_end]))
+
+    #appended_data = Vector{UInt8}(rstrip(raw_file_contents[offset_begin:offset_end]))
+    # rstrip was used earlier, but removes a space at the end of the line, which can cause havoc 
+    # in case raw_file_contents ends with a space, such as 
+    # "... (f\x12 \n\t", which will be turned into "... (f\x12" 
+    appended_data = Vector{UInt8}(raw_file_contents[offset_begin:offset_end][1:(end - 2)])
```
sloede, line 104:
```suggestion
    appended_data = Vector{UInt8}(raw_file_contents[offset_begin:offset_end])
```
I think this is what we should actually do here. `rstrip` is too eager and was based on the VTK wiki, which [states](https://vtk.org/Wiki/VTK_XML_Formats#Appended_Data_Section)
> Extra whitespace AFTER the data has no effect.

I originally and mistakenly interpreted this as "extra whitespace at the end can be stripped". However, I now believe this really means "if there is extra whitespace after the actual data, it will not affect the correct reading of the data".

IMHO this makes sense, since after the initial data position is identified (i.e., after the `_`), all data is read with an offset and a length, thus any extra data at the end will never be read.

Therefore, I propose that we just remove the `rstrip` completely and rely on the fact that data is never read just "to the end of the data array" but always with a fixed length. Maybe this will also get rid of the issue that sometimes the compressed data was shorter than the supposed length, which you encountered in #18 and for which you added a [check](https://github.com/JuliaVTK/ReadVTK.jl/blob/3c0cdb8495ef14155ce7971d8517f1c1440073bd/src/ReadVTK.jl#L732-L737).

boriskaus: Yes that might do the job too. I have a test file for this to check (that is not too large in size) but wasn’t sure whether you are ok that this is uploaded as part of the current repo.

boriskaus:
I can confirm that the suggested changes work & all tests pass fine. A test file is now uploaded above; feel free to add it to this repo or to a separate one.

sloede:
> A test file is now uploaded above; feel free to add it to this repo or to a separate one.

I don't see the test file. However, we have a separate repository for that, https://github.com/JuliaVTK/ReadVTK_examples. It would be great if you can create a PR to that repo to add the new test file there, then we can use it for a new test here.

boriskaus:
> It would be great if you can create a PR to that repo to add the new test file there, then we can use it for a new test here.

done [here](https://github.com/JuliaVTK/ReadVTK_examples/pull/2).

sloede:
> done [here](https://github.com/JuliaVTK/ReadVTK_examples/pull/2).

Thanks! I merged it so you can go ahead and add a test here.

boriskaus:
done, have a look if you're happy

test/rectilinear.jl
```diff
@@ -133,3 +133,9 @@ for compress in [true, false]
 
   end
 end
+
+# Testfile with trailing whitespace
```
sloede, line 137: Could you please put the entire new test in its own `@testset "trailing whitespace in appended data" begin … end` block? That way it is easier to figure out which test failed from the test overview. I’m on a mobile device without the ability to do multi line comments, thus I can’t make this a proper suggestion, sorry.

sloede (approved):
LGTM! Thanks for the bug fix!

boriskaus:
Thanks - can you register the new version so I can update my packages accordingly?

sloede:
Yes, I already tried to do it but we first need https://github.com/JuliaRegistries/General/pull/83019 to be merged, since we haven't changed the repo URL after moving this package to the JuliaVTK org. I'll let you know as soon as v0.1.7 is registered.

sloede:
The new release v0.1.7 is out. Sorry for the delay!

boriskaus:
Thanks a lot!

PR #42, opened 2023-09-28
get_field(vtk) function added to get FieldData from VTK

andresegido:
Introduced functions 'get_field' and 'field_data' in order to be able to read FieldData from VTK files.

### DIFF
--- a/src/ReadVTK.jl
+++ b/src/ReadVTK.jl
@@ -184,6 +184,11 @@ function piece(vtk_file::VTKFile)
   return LightXML.root(vtk_file.xml_file)[vtk_file.file_type][1]["Piece"][1]
 end
 
+# Return `FieldData` XML element that contains all VTK field data
+function field_data(vtk_file::VTKFile)
+  return LightXML.root(vtk_file.xml_file)[vtk_file.file_type][1]["FieldData"][1]
+end
+
 """
     isstructured(xml_file)
 
@@ -406,6 +411,23 @@ function get_data_section(vtk_file::VTKFile, section)
   return VTKData(names, data_arrays, vtk_file)
 end
 
+# Retrieve FieldData section from the VTK file
+function get_field(vtk_file::VTKFile)
+  names = String[]
+  data_arrays = XMLElement[]
+
+  # Iterate over XML elemens in the section
+  for xml_element in child_elements(field_data(vtk_file))
+    # We do not know how to handle anything other than `DataArray`s
+    @assert LightXML.name(xml_element) == "DataArray"
+
+    # Store the name and the XML element for each found data array
+    push!(names, attribute(xml_element, "Name", required = true))
+    push!(data_arrays, xml_element)
+  end
+
+  return VTKData(names, data_arrays, vtk_file)
+end
 
 """
     get_cell_data(vtk_file::VTKFile)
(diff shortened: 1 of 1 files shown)


src/ReadVTK.jl
```diff
@@ -406,6 +411,23 @@ function get_data_section(vtk_file::VTKFile, section)
   return VTKData(names, data_arrays, vtk_file)
 end
 
+# Retrieve FieldData section from the VTK file
```
sloede, line 414: This should get a docstring

src/ReadVTK.jl
```diff
@@ -406,6 +411,23 @@ function get_data_section(vtk_file::VTKFile, section)
...
+function get_field(vtk_file::VTKFile)
```
sloede, line 415:
```suggestion
function get_field_data(vtk_file::VTKFile)
```
for consistency with `get_cell_data` and `get_point_data`?

sloede (changes requested):
Thanks a lot for this contribution 🙏! This PR seems to be on the right track. There are some usability issues open (see the comments), and it would be great if you could add some simple tests for the new functionality.

andresegido:
Thanks for the suggestions! I just introduced them into my last commit.

src/ReadVTK.jl
```diff
@@ -411,8 +411,14 @@ function get_data_section(vtk_file::VTKFile, section)
   return VTKData(names, data_arrays, vtk_file)
 end
 
-# Retrieve FieldData section from the VTK file
-function get_field(vtk_file::VTKFile)
+"""
+  get_field_data(vtk_file::VTKFile)
```
ranocha, line 415:
```suggestion
    get_field_data(vtk_file::VTKFile)
```
4 spaces are needed (or typically used) to render this as code

ranocha:
> Thanks a lot for this contribution 🙏! This PR seems to be on the right track. There are some usability issues open (see the comments), and it would be great if you could add some simple tests for the new functionality.

Thanks a lot also from my side! Could you please add some tests of the new functionality? See `test/runtests.jl`

andresegido:
> > Thanks a lot for this contribution 🙏! This PR seems to be on the right track. There are some usability issues open (see the comments), and it would be great if you could add some simple tests for the new functionality.
> 
> Thanks a lot also from my side! Could you please add some tests of the new functionality? See `test/runtests.jl`

Hi! I don't have much experiences writing tests, and much less in Julia; so I'm not sure about where to start...

Also, I did a quick check through the vtk files you use for your tests and none of them have any field data, so using them would not be desirable for such a test.

I have done some work with a dummy vtk file I have locally and it works as expected:

"""
vtk = VTKFile("testing/Wing.vtu")
fieldData = ReadVTK.get_field_data(vtk)
nx = get_data(fieldData["nx"])
"""

In this case, fieldData behaves as a ReadVTK.VTKData and nx reads the expected value.

This might help you appending some field data to one of your vtk files and preparing a test.

ranocha:
The example files are collected in the repository https://github.com/JuliaVTK/ReadVTK_examples. You can open a PR there and upload a new file containing some field data. When that PR is merged, you can update the commit hash in
https://github.com/JuliaVTK/ReadVTK.jl/blob/7668d43a85918756719d62bf95e1085cdf539304/test/runtests.jl#L5-L9
accordingly and write a proper test here.

Issue #43, opened 2023-10-07
Constraint on the VTK version


henry2004y:
Hi,

While testing to read one ImageData generated by ParaView 5.11, I noticed that the VTK file version was asserted to be `v"1.0"`. The file version generated by ParaView 5.11 is 2.2, and the latest VTK version is 3.0.

I am wondering if there is a reason for only supporting version 1.0? If I remove

```julia
@assert version == v"1.0"
```

it seems that the reader still works in my limited attempts.

Thanks!

sloede:
I checked and indeed, there is no official VTK XML spec file findable on the internet. We thus do not know what version 1 or higher (or lower) really means. Thus, feel free to create a PR that removes this check and we will be happy to merge it.

henry2004y:
I know nothing about the specific differences, but let's hope it doesn't break anything.

The [VTKFileFormats](https://docs.vtk.org/en/latest/design_documents/VTKFileFormats.html) briefly mentioned that `v1.0` and `v2.0` is compatible with `v3.0`, and some older names are discouraged (e.g. aspect ratio in `v1.0` v.s. spacing in `v2.0` for structured points). I do not see any real difference.

sloede:
> The [VTKFileFormats](https://docs.vtk.org/en/latest/design_documents/VTKFileFormats.html) briefly mentioned that `v1.0` and `v2.0` is compatible with `v3.0`, and some older names are discouraged (e.g. aspect ratio in `v1.0` v.s. spacing in `v2.0` for structured points). I do not see any real difference.

Yes, but they refer to the legacy formats, while we only support the XML file formats, for which no version information has ever been released publicly (see also https://gitlab.kitware.com/vtk/vtk/-/issues/17675).

MarDiehl:
I use a rather recent version of VTK (9.3.0 from https://archlinux.org/packages/extra/x86_64/vtk) and the written image data has version 0.1. Would it be possible to further relax the version and essentially accept everything until the different versions are documented?

sloede:
Yes, that's fine with me. If you create a PR that removes the checks, I'll merge it 😊

PR #44, opened 2023-10-07
Loose VTK XML version constraint

henry2004y:
This PR attempts to loose the current constraint on the VTK XML version requirement, based on the discussion in #43.

### DIFF
--- a/src/ReadVTK.jl
+++ b/src/ReadVTK.jl
@@ -28,7 +28,7 @@ Hold all relevant information about a VTK XML file that has been read in.
 - `filename`: original path to the VTK file that has been read in
 - `xml_file`: object that represents the XML file
 - `file_type`: currently only `"UnstructuredGrid"` or `"ImageData"` are supported
-- `version`: currently only v1.0 is supported
+- `version`: currently only v1.0 and beyond are supported
 - `byte_order`: can be `LittleEndian` or `BigEndian` and must currently be the same as the system's
 - `compressor`: can be empty (no compression) or `vtkZLibDataCompressor`
 - `appended_data`: in case of appended data (see XML documentation), the data is stored here for
@@ -125,7 +125,7 @@ function VTKFile(filename)
   end
 
   # Ensure correct version
-  @assert version == v"1.0"
+  @assert version ≥ v"1.0"
 
   # Ensure matching byte order
   is_little_endian = ENDIAN_BOM == 0x04030201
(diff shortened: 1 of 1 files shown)


src/ReadVTK.jl
```diff
@@ -28,7 +28,7 @@ Hold all relevant information about a VTK XML file that has been read in.
 - `filename`: original path to the VTK file that has been read in
 - `xml_file`: object that represents the XML file
 - `file_type`: currently only `"UnstructuredGrid"` or `"ImageData"` are supported
-- `version`: currently only v1.0 is supported
+- `version`: currently only v1.0 and beyond are supported
```
sloede, line 31:
```suggestion
- `version`: VTK XML file format version (v1.0 or later)
```
I think there is no useful spec information on this available publicly, so we can just be very neutral about this

src/ReadVTK.jl
```diff
@@ -213,7 +213,7 @@ Hold all relevant information about a Parallel VTK XML file that has been read i
 - `filename`: original path to the PVTK file that has been read in
 - `xml_file`: xml info
 - `file_type`: currently only `"PRectilinearGrid"` or `"PImageData"` are supported
-- `version`: currently only v1.0 is supported
+- `version`: currently only v1.0 and beyond are supported
```
sloede, line 216:
```suggestion
- `version`: VTK XML file format version (v1.0 or later)
```
as above

src/ReadVTK.jl
```diff
@@ -258,7 +258,7 @@ function PVTKFile(filename; dir = "")
   end
 
   # Ensure correct version
-  @assert version == v"1.0"
+  @assert version ≥ v"1.0"
```
sloede, line 261:
```suggestion
  @assert version >= v"1.0"
```
as above

src/ReadVTK.jl
```diff
@@ -323,7 +323,7 @@ function PVDFile(filename)
...
```
sloede, line 326:
```suggestion
  @assert version >= v"1.0"
```
as above

src/ReadVTK.jl
```diff
@@ -125,7 +125,7 @@ function VTKFile(filename)
...
```
sloede, line 128:
```suggestion
  @assert version >= v"1.0"
```
We prefer to only use ASCII characters for code

sloede (changes requested):
Thanks a lot for your contribution! With some minor changes, this LGTM.

src/ReadVTK.jl
henry2004y: Yep

sloede (approved):

sloede:
v0.1.9 has just been released with your fix included.

PR #47, opened 2023-11-08
Add 1 to `VTKPrimitives` connectivity indices to follow convention in ReadVTK.jl

eliascarv:
This convention is already working in the ".vtu" format and needs to be corrected in the ".vtp" format

### DIFF
--- a/src/ReadVTK.jl
+++ b/src/ReadVTK.jl
@@ -1069,7 +1069,8 @@ function get_cells(vtk_file::VTKFile)
   # Create VTKCells container and convert VTK's zero-based indices to Julia's one-based indices
   # Note: the offsets do not need to be updated since they point *past* the last entry
   # in VTK files (C-style), while in Julia it is custom to point *at* the last entry.
-  return VTKCells(get_data(connectivity) + oneunit.(get_data(connectivity)),
+  connectivity_data = get_data(connectivity)
+  return VTKCells(connectivity_data .+ oneunit(eltype(connectivity_data)),
                   get_data(offsets),
                   get_data(types))
 end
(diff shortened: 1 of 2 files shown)


src/ReadVTK.jl
```diff
@@ -1069,7 +1069,8 @@ function get_cells(vtk_file::VTKFile)
   # Create VTKCells container and convert VTK's zero-based indices to Julia's one-based indices
   # Note: the offsets do not need to be updated since they point *past* the last entry
   # in VTK files (C-style), while in Julia it is custom to point *at* the last entry.
-  return VTKCells(get_data(connectivity) + oneunit.(get_data(connectivity)),
+  connectivity_data = get_data(connectivity)
+  return VTKCells(connectivity_data .+ oneunit(eltype(connectivity_data)),
```
sloede, line 1073: What's the effect of this change, i.e., why was it rewritten like this (honest question)?

sloede (commented):
Thanks for the contribution! I have just one minor question.

@ranocha I think this will require a minor version bump since the observable behavior changes, what do you think?

src/ReadVTK.jl
eliascarv: The way it was being done before, an array of 1s was allocated to add to connectivity_data. 
I noticed that the `VTKDataArray{T}` will always contain elements of the same type.
Then I realized I could use broadcast to add `oneunit(T)` to all elements and save an allocation.

sloede: Sounds reasonable 👍

sloede (approved):
LGTM, thanks!

ranocha (approved):
Thanks!

@sloede Feel free to go ahead as you proposed 👍

sloede:
ReadVTK.jl v0.2.0 was just released, which includes this PR.

PR #61, opened 2024-08-19
Update/add show functions VTK objects

williamjsdavis:
The following pull request updates and adds `show` methods for various VTK objects, following this issue: https://github.com/JuliaVTK/ReadVTK.jl/issues/60

Updated and added methods are given for the following objects: 
- `VTKFile`,
- `VTKDataArray`,
- `PVTKFile`, 
- `PVDFile`, 
- `VTKData`, 
- `PVTKData`, and 
- `PVTKDataArray`.

Where possible, I have tried to copy the style of the original `show` method defined for `VTKFile` objects.

The following subsections show a comparison of the original `show` output verses the proposed new `show` output (generated from files in the test suite).

#### Original output
```
VTKFile("./celldata_appended_binary_compressed.vtu", <XMLDocument>, "UnstructuredGrid", "1.0.0", "LittleEndian", "vtkZLibDataCompressor", <appended_data>, 4434, 3085)
```

#### Proposed new output
```
VTKFile("./celldata_appended_binary_compressed.vtu", <XMLDocument>, "UnstructuredGrid", v"1.0.0", "LittleEndian", "vtkZLibDataCompressor", <appended_data>, 4434, 3085)
```

#### Notes

The proposed changes are only slightly different than the current implementation. The only difference is using the `repr` representation for the `vtk_file.version::VersionNumber`.

#### Original output
```
VTKDataArray("element_ids")
```

#### Proposed new output
```
VTKDataArray{Int64, 1, ReadVTK.FormatAppended}("element_ids", 32675, <DataArray type="Int64" Name="element_ids" NumberOfComponents="1" format="appended" offset="32675"/>, <VTKFile>)
```

#### Notes

The new `show` method lists the parametric types associated with the object, listing the data type encoding, the size of the second dimension, and the encodingformat`. It also lists the other data fields, with a shortened expression for the `vtk_file` field (shortened similarly to the method defined for `VTKFile`). 

#### Original output
```
PVTKFile()
```

#### Proposed new output
```
PVTKFile("fields.pvtr", <XMLDocument>, "PRectilinearGrid", v"1.0.0", ["fields/fields_1.vtr", "fields/fields_2.vtr", "fields/fields_3.vtr", "fields/fields_4.vtr"], <vtk_files>)
```

#### Notes

Similar format to `VTKFile`, but including a list representation for the field `vtk_filenames::Vector{String}`.

#### Original output
```
PVDFile()
```

#### Proposed new output
```
[...]

PR #62, opened 2024-08-19
fix tests for WriteVTK.jl v1.19 and run new version of the formatter

ranocha:
Since #61 failed, I wanted to test CI here and also ran the newest version of the formatter.


ranocha:
Looks like the problem comes from using relative paths instead of absolute paths in PVTK files - xref https://github.com/JuliaVTK/WriteVTK.jl/pull/141

PR #65, opened 2024-09-03
Skip Array data when constructing VTKData

williamjsdavis:
Regarding issue https://github.com/JuliaVTK/ReadVTK.jl/issues/63, when reading data from a vtk/pvtk file, the current read implementation only work if all elements in the `PointData`/`CellData` section are of type `DataArray`. In some use cases, string data is stored in these sections as `Array` type, (e.g., as label for visualization projects, [link](https://vtk.org/doc/nightly/html/classvtkStringArray.html)). These elements have the form:
<Array type="String" Name="Container" format="ascii">

Currently, if there is a single `Array` element in the `PointData`/`CellData` section, reading data will fail due to an assertion breach in the `get_data_section` method:

https://github.com/JuliaVTK/ReadVTK.jl/blob/c9144477b4c8a9bf41bf0cbab9518d0a2e86821a/src/ReadVTK.jl#L425

In the new implementation I present in this pull request, XML elements that are not `DataArray` type are skipped over in `get_data_section`, but error later when one attempts to index the `VTKData` object with a key that references non-numeric data: 

https://github.com/JuliaVTK/ReadVTK.jl/blob/c9144477b4c8a9bf41bf0cbab9518d0a2e86821a/src/ReadVTK.jl#L530

This new logic allows for the reading of numeric data from files with both numeric and non-numeric contents.

I have also expanded existing tests to:
1. Add string data to a generated `.vti` file,
2. Add string data to a generated `.pvti` file,
3. Test for expected `KeyError`s when attempting to read the string data for the `.vti` file, and
4. Test for expected `KeyError`s when attempting to read the string data for the `.pvti` file

And the fact that the existing tests for reading cell/point data pass as they previously were doing indicates that numeric data can still be correctly read from `.vti` and `.pvti` files.

Feedback is appreciated!

### DIFF
--- a/README.md
+++ b/README.md
@@ -125,6 +125,8 @@ Further contributions to ReadVTK have been made by the following people:
   (Johannes-Gutenberg University Mainz, Germany)
 * [Matthew Whisenant](https://volweb2.utk.edu/~mwhisena/)
   (University of Tennessee, Knoxville)
+* [William Davis](https://posgeo.wordpress.com/)
+  (Terra AI, USA)
 
 ## License and contributing
 ReadVTK is licensed under the MIT license (see [LICENSE.md](LICENSE.md)).
--- a/src/ReadVTK.jl
+++ b/src/ReadVTK.jl
@@ -415,6 +415,7 @@ end
 
 
 # Retrieve a data section (should be `CellData` or `PointData`) from the VTK file
+
 function get_data_section(vtk_file::VTKFile, section)
   names = String[]
   data_arrays = XMLElement[]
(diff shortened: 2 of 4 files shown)


ranocha:
Thanks! This looks quite good to me. Would you like to add your name to the list of authors in https://github.com/JuliaVTK/ReadVTK.jl?tab=readme-ov-file#authors?

williamjsdavis:
Yes please! I'd be very happy to continue contributing! Please add me as: 

[William Davis](https://posgeo.wordpress.com/) (Terra AI, USA)

ranocha:
Could you please do that in this PR?

williamjsdavis:
Ah understood! I've added it to the PR!

ranocha (approved):
Thanks

PR #70, opened 2025-07-24
Pass `XML_PARSE_HUGE` to `LightXML.parse_string`

tverho:
New versions of libxml2 have limits for the size of XML files, which are sometimes exceeded with large VTK meshes. The limitations can be relaxed with the option `XML_PARSE_HUGE`. After https://github.com/JuliaIO/LightXML.jl/commit/eed623754562f89392ae922525c8fcf2f9746f82 LightXML allows passing options to `parse_string`, and this pull request makes use of that to include the option `XML_PARSE_HUGE`.


ranocha (approved):
Thanks

